PR #8, opened 2022-08-05
Socket connections params support for Android and iOS

whoiswentz:


README.md
```diff
@@ -139,6 +139,16 @@ On iOS installation is a bit more complicated, because you need to setup a scree
...
+   > This new dependency should be added outside of your application target. Example
+   > ```rb
+   > target 'ReactNativeMembraneExample' do
+   >  ...
+   > end
+   > 
+   > target 'MembraneScreenBroadcastExtension' do
+   >  pod 'MembraneRTC/Broadcast'
+   > end
+   > ```
```
graszka22, line 151: Weird, it's already merged and it showed here, did you start with updated master branch? There were a lot of changes in the last commit.

example/package-lock.json
```diff
@@ -0,0 +1,16219 @@
+{
```
graszka22, line 1: we're using yarn for development, please remove package-lock

graszka22: also there are a lot of changes in `yarn.lock` and I don't get why, maybe because you've used npm?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@membraneframework/react-native-membrane-webrtc",
-  "version": "1.0.1",
+  "version": "1.2.0",
```
graszka22, line 3: don't bump this version, we'll bump it when we'll make a release

android/src/main/java/com/reactnativemembrane/MembraneModule.kt
```diff
@@ -90,20 +90,37 @@ class MembraneModule(reactContext: ReactApplicationContext) :
     return res
   }
 
+  private fun ReadableMap.toMap(): MutableMap<String, Any> {
```
graszka22, line 93: Do we need all of those booleans, numbers etc? Shouldn't strings suffice? If so, we have already function for that: `ReadableMap.toMetadata`

graszka22 (changes requested):
Thanks for contributing! I added a couple of comments.

Also please update documentation in README about those connection params

README.md
whoiswentz: Yes, I forked the project and after I pull the latest changes from master, that's really weird

example/package-lock.json
whoiswentz: yes, I'm used to use npm, I will remove all the packages-lock.json

graszka22 (approved):

PR #12, opened 2022-08-26
Add update metadata api

graszka22:


mickel8 (approved):
Honestly speaking I know nothing about react but the API looks good :+1:

PR #14, opened 2022-10-05
[MV-138] [MV-170] Add simulcast and max bandwidth API

graszka22:


README.md
```diff
@@ -416,6 +428,55 @@ Props:
...
+## `useSimulcast()`
+
+This hook manages the simulcast configuration of a video track.
+
+### Returns
+
+An object containing:
+
+- `simulcastConfig` - current [SimulcastConfig](#SimulcastConfig) of a video track
+- `selectReceivedTrackTargetEncoding(peerId: string, encoding: TrackEncoding)` - selects track encoding that server should send to the client library. The encoding will be sent whenever it is available. If choosen encoding is temporarily unavailable, some other encoding will be sent until choosen encoding becomes active again. Arguments:
```
mickel8, line 440: I think that in other SDKs we have 

```suggestion
- `setTargetTrackEncoding(trackId: string, encoding: TrackEncoding)` - sets track encoding that server should send to the client library. The encoding will be sent whenever it is available. If choosen encoding is temporarily unavailable, some other encoding will be sent until choosen encoding becomes active again. Arguments:
```

mickel8 (approved):

PR #17, opened 2022-10-17
[MV-175] Add online docs

graszka22:
Docs available on: https://docs.membrane.stream/react-native-membrane-webrtc/


mickel8 (approved):
As in the case of android, could we also link the documentation at the beginning of README?

Also, looks like something is not linked correctly 

image

PR #44, opened 2023-03-08
[MV-336]  Add basic text and color styling.

skyman503:
This change introduces basic text and color styling that will be used for all components in this project. Followup PR will apply styling to existing components.


example/android/link-assets-manifest.json
```diff
@@ -0,0 +1,9 @@
+{
```
graszka22, line 1: Have your tried using expo's google fonts? https://docs.expo.dev/guides/using-custom-fonts/#using-a-google-font
Inter font is even shown in the example.
That way there is no need for including ttf in native side, it should be easier.

example/src/App.tsx
```diff
@@ -113,33 +114,33 @@ export default function App() {
 
   return (
     <View style={styles.container}>
-      <Text>Room name:</Text>
+      <Text style={Fonts.Headlines.h5}>Room name:</Text>
```
graszka22, line 117: I'd prefer to have our custom styled `<Text>` component with a prop like this:
`<Text variant="h5">Room name:</Text>`
Or you can name it `<Typo>` to make it less confusing with standard RN Text. 
That way it's less typing and no need to include styles and when we want to change all texts in the app we'd need to do it in one place.
Also make it typed so that vscode will suggest and check text variant.
Other components can be done in a similar way.

graszka22 (changes requested):

example/android/app/build.gradle
```diff
@@ -144,12 +144,12 @@ android {
         }
     }
     signingConfigs {
-        release {
-            storeFile file(MEMBRANE_UPLOAD_STORE_FILE)
-            storePassword MEMBRANE_UPLOAD_STORE_PASSWORD
-            keyAlias MEMBRANE_UPLOAD_KEY_ALIAS
-            keyPassword MEMBRANE_UPLOAD_KEY_PASSWORD
-        }
+//        release {
```
graszka22, line 147: You commited this by accident.
I'll fix that now so that you don't need to comment that out.

example/android/link-assets-manifest.json
skyman503: I thought about it but that requires me to load the font using useFonts() in any component, and I wasn't sure where to load it. I tried doing that in the main App.tsx component but that resulted with a bunch of errors since the font wasn't loaded in time to be used for style declarations in the fonts/fonts.ts, which is independed from the App.tsx.

However if you have any tips on how to load a font when it is being used outside of the react components, I'd love to try it out.

example/src/App.tsx
skyman503: SG, I'll do that.

example/android/link-assets-manifest.json
graszka22: Sure! In another project we use a splash screen (`expo-splash-screen` and `preventAutoHideAsync` function). So while we load the font and other assets we just show the splashscreen and after loading we show the app with `expo-splash-screen`'s `hideAsync`. We have the splashscreen anyway in the designs.

graszka22: You have this even described here: https://docs.expo.dev/guides/using-custom-fonts/#waiting-for-fonts-to-load

example/src/App.tsx
```diff
@@ -15,6 +17,9 @@ import {
 } from 'react-native';
 
 import { Room } from './Room';
+import { Typo } from './fonts/Typo';
```
graszka22, line 20: I'd put the Typo and  Icon components in a folder named `components`. Then we'd import it like this: `import { Typo } from '@components/Typo'`. Then we'll put other components here. 
Use https://www.npmjs.com/package/babel-plugin-module-resolver for those nice imports.

example/src/fonts/Typo.tsx
```diff
@@ -0,0 +1,235 @@
...
+  caption: {
+    fontFamily: 'Inter',
+    fontSize: 18,
+    fontWeight: 'bold',
+    lineHeight: 24,
+    letterSpacing: 1,
+    textTransform: 'uppercase',
+  },
+  button: {
+    fontFamily: 'Inter-Medium',
```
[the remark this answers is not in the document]
skyman503: Changed all to be `Inter-Medium` since this is only temporary font, I don't think there is a point in additional hustle.

example/src/fonts/Typo.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+export const Typo = (props: TypoProps) => {
+  const windowDimensions = Dimensions.get('window');
+
+  const [dimensions, setDimensions] = useState({
+    window: windowDimensions,
+  });
+
+  useEffect(() => {
+    const subscription = Dimensions.addEventListener(
```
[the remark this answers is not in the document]
skyman503: Good point, thanks!

example/src/App.tsx
skyman503: Done.

example/src/icons/index.ts
```diff
@@ -1,9 +1 @@
-export { default as CameraIcon } from './outline-cam.svg';
-export { default as CameraDisabledIcon } from './outline-cam-disabled.svg';
-export { default as MicrophoneDisabledIcon } from './outline-microphone-disabled.svg';
-export { default as MicrophoneIcon } from './outline-microphone.svg';
-export { default as PhoneDownIcon } from './outline-phone-down.svg';
 export { default as FlipCameraIcon } from './outline-refresh.svg';
```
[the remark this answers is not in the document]
graszka22: If it's not on the designs you can get rid of the icon and the button completely, I think that would be the easiest way

example/src/fonts/Typo.tsx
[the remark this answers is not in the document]
graszka22: What about the second part? The hook still subscribes to events under the hood. Also the breakpoints are on height or width of the screen?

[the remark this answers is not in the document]
skyman503: Sorry, I misunderstood your intention. Should be good now.

graszka22 (commented):
Have you tested it on Android? I added some missing gradle code so that it would build on my machine. Please test it on both systems next time.
Also you haven't responded: is the breakpoint on the height or width of the screen? Currently it's coded as width, but the design doesn't specify it.

example/src/fonts/Typo.tsx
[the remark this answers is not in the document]
skyman503: I just assumed that width is better way to determine when to shrink the font size since text is generally wider than taller. I sent a question about that to the designer, but she is OOO until later today.

graszka22 (approved):
Ok, LGTM now 👏 You can merge it now and eventually change height to width when you get response from the designer.

PR #48, opened 2023-03-09
[MV-336] Add in call buttons

skyman503:
This changes introduces custom button component for all buttons that user can interact with during a call.


example/src/components/buttons/InCallButton.tsx
```diff
@@ -0,0 +1,63 @@
+import React, { ReactNode } from 'react';
+import { StyleSheet, View, Pressable } from 'react-native';
+
+import { AdditionalColors, BrandColors } from '../../shared/colors';
```
graszka22, line 4: You can make `@colors` alias

example/src/components/buttons/InCallButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  primary: {
+    borderWidth: 1,
+    borderColor: BrandColors.darkBlue80,
+    borderStyle: 'solid',
+    backgroundColor: AdditionalColors.white,
+  },
+  danger: {
```
graszka22, line 22: If it's for disconnect button it's not actually "dangerous". By "dangerous" I've meant buttons like this one:
Screenshot 2023-03-09 at 15 24 02
Maybe it'd be better to name it "disconnect" or something? Also it's got a frame like a primary button.

example/src/components/buttons/InCallButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+        return BrandColors.darkBlue100;
+      case 'danger':
+        return AdditionalColors.white;
+    }
+  };
+
+  return (
+    <Pressable onPress={() => {}}>
+      <View style={GetStylesForButtonType(props.type)}>
+        {React.cloneElement(props.children as React.ReactElement<any>, {
```
graszka22, line 56: Why can't we just pass icon name as a prop and use Icon component here? Wouldn't it be much simpler?

graszka22 (changes requested):
Maybe it would be a good idea to use those buttons already in the app in order to test them? Changing `Room.tsx` is ok, it's going to be rewritten soon anyway.

example/src/components/buttons/InCallButton.tsx
skyman503: That requires developers later on to always maintain the up-to-date list of all possible icons in this file, which I wasn't sure if it's a good practice. If you prefer I can do that.

graszka22: Icon name can be just a string. In fact, the Icon component itself has a prop `name` that's just a string in case of icomoon icons. What a bummer. Maybe there is a clever way to do it, but for now let's make it just a string.

example/src/components/buttons/InCallButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+    switch (type) {
+      case 'primary':
+        return BrandColors.darkBlue100;
+      case 'danger':
+        return AdditionalColors.white;
+    }
+  };
+
+  return (
+    <Pressable onPress={() => {}}>
```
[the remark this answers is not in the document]
skyman503: Done.

example/src/components/buttons/InCallButton.tsx
skyman503: Done.

example/src/components/buttons/InCallButton.tsx
skyman503: Done.

example/src/components/buttons/InCallButton.tsx
skyman503: I didn't want to call it like that, since in the future this might be misunderstood  since as the app get bigger we might want to reuse this styling for something different.

Also I don't think I understand WDYM by "a frame like primary button". I extracted the common part out of their respective stylings.

example/src/Controls.tsx
```diff
@@ -28,43 +26,60 @@ export const Controls = ({ disconnect }: { disconnect: () => void }) => {
...
+      ) : (
+        <InCallButton
+          type="primary"
+          iconName="Microphone"
+          onPress={toggleMicrophoneAndUpdateMetadata}
+        />
+      )}
+      {!isCameraOn ? (
+        <InCallButton
+          type="primary"
```
graszka22, line 57: we can make the primary one default = less typing

example/src/components/buttons/InCallButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+        return BrandColors.darkBlue100;
+      case 'disconnect':
+        return AdditionalColors.white;
+    }
+  };
+
+  return (
+    <Pressable
+      onPress={() => {
+        props.onPress();
```
graszka22, line 57: You can just pass `onPress={props.onPress}`?
Also you can use destructuring syntax like this:
```
export const InCallButton = ({
  type,
  onPress,
  iconName,
}: InCallButtonProps) => {
```
and just use `onPress={onPress}` but whatever you prefer

skyman503: Done.

example/src/Controls.tsx
skyman503: Done.

example/src/components/buttons/InCallButton.tsx
skyman503: Changed to disconnect.

graszka22 (approved):

PR #49, opened 2023-03-10
[MV-338] Add react-navigation

skyman503:
This changes add react-navigation library and use it to navigate between home screen and room screen.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,45 +1,18 @@
-import { Typo } from '@components/Typo';
-import { SERVER_URL } from '@env';
-import * as Membrane from '@jellyfish-dev/react-native-membrane-webrtc';
+import { NavigationContainer } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
 import { useFonts } from 'expo-font';
 import * as SplashScreen from 'expo-splash-screen';
-import React, { useCallback, useEffect, useState } from 'react';
-import {
-  StyleSheet,
-  View,
-  PermissionsAndroid,
-  SafeAreaView,
-  Text,
-  TextInput,
-  Platform,
-  Pressable,
-  Switch,
-  Alert,
-} from 'react-native';
+import React, { useEffect } from 'react';
 
+import { Home } from './Home';
 import { Room } from './Room';
+import { VideoroomContextProvider } from './VideoroomContext';
 
 SplashScreen.preventAutoHideAsync();
 
+const Stack = createNativeStackNavigator();
+
 export default function App() {
-  const {
-    connect: mbConnect,
-    disconnect: mbDisconnect,
-    joinRoom,
-    error,
-  } = Membrane.useMembraneServer();
-  const [connected, setConnected] = useState<boolean>(false);
-  const [roomName, setRoomName] = useState<string>('room');
-  const [serverUrl, setServerUrl] = useState<string>(SERVER_URL);
-  const [displayName, setDisplayName] = useState<string>(
-    `mobile ${Platform.OS}`
-  );
-  const [isSimulcastOn, setIsSimulcastOn] = useState<boolean>(true);
-
-  const params = {
-    token: 'NOW_YOU_CAN_SEND_PARAMS',
-  };
-
   const [fontsLoaded] = useFonts({
     'Inter-Medium': require('../assets/fonts/Inter-Medium.ttf'),
     'IcoMoon': require('../assets/fonts/icomoon/icomoon.ttf'),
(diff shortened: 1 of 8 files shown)


example/src/Room.tsx
```diff
@@ -1,12 +1,12 @@
...
 import * as Membrane from '@jellyfish-dev/react-native-membrane-webrtc';
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import { StyleSheet, Text, View, Pressable } from 'react-native';
 
 import { Controls } from './Controls';
 import { Settings } from './Settings';
 
-export const Room = ({ disconnect }: { disconnect: () => void }) => {
+export const Room = ({ navigation }) => {
```
graszka22, line 9: The back button should disconnect the call or there shouldn't be any back button. Now you can go back and connect again and that leads to weird behaviour. I can see that according to the designs there shouldn't be even a header. You can change that in navigator options https://reactnavigation.org/docs/native-stack-navigator#options

example/src/VideoroomContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const [serverUrl, setServerUrl] = useState<string>(SERVER_URL);
+  const [displayName, setDisplayName] = useState<string>(
+    `mobile ${Platform.OS}`
+  );
+  const [isSimulcastOn, setIsSimulcastOn] = useState<boolean>(true);
+
+  return (
+    <VideoroomContext.Provider
+      value={{
+        roomNameState: [roomName, setRoomName],
```
graszka22, line 28: Wouldn't it be easier to do something like that:
```
value = {{
  roomName,
  setRoomName,
  serverUrl,
  setServerUrl,
  ...
}}
```
?

example/src/Home.tsx
```diff
@@ -0,0 +1,160 @@
...
+  PermissionsAndroid,
+  Switch,
+  Alert,
+} from 'react-native';
+
+import { VideoroomContext } from './VideoroomContext';
+
+export const Home = ({ navigation }) => {
+  const videoroomContext = useContext(VideoroomContext);
+  const [roomName, setRoomName] = videoroomContext!.roomNameState;
```
graszka22, line 20: Why do we need `!` here? Have you looked into the link I attached in the task description? It has a solution to exactly this problem: https://kentcdodds.com/blog/how-to-use-react-context-effectively#typescript

example/src/VideoroomContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+  undefined
+);
+
+const VideoroomContextProvider = (props) => {
+  const [roomName, setRoomName] = useState<string>('room');
+  const [serverUrl, setServerUrl] = useState<string>(SERVER_URL);
+  const [displayName, setDisplayName] = useState<string>(
+    `mobile ${Platform.OS}`
+  );
+  const [isSimulcastOn, setIsSimulcastOn] = useState<boolean>(true);
```
graszka22, line 23: Do we even need this in the context? We're not using it anywhere besides Home screen. This applies also to `serverUrl`. In the context we should only store things that are needed globally in the app.

example/src/App.tsx
```diff
@@ -1,45 +1,18 @@
...
-  Text,
-  TextInput,
-  Platform,
-  Pressable,
-  Switch,
-  Alert,
-} from 'react-native';
+import React, { useEffect } from 'react';
 
+import { Home } from './Home';
```
graszka22, line 7: Consider putting screens in a separate folder `screens` and maybe importing them like `@screens`? We're going to have a bit more of screens in the future.

graszka22 (changes requested):

example/src/App.tsx
skyman503: Done.

example/src/Home.tsx
skyman503: Yes I read it, I just thought it would look silly with the amount of values I wanted to store there.  After other changes I think it makes sens now. Done.

example/src/VideoroomContext.tsx
skyman503: Moved to `Home.tsx`

example/src/Room.tsx
skyman503: Done.

example/src/VideoroomContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+  switch (action.type) {
+    case 'set':
+      return { roomName: action.val };
+    default:
+      throw new Error(`Unhandled action type: ${action.type}`);
+  }
+}
+
+const VideoroomContextProvider = (props) => {
+  const [state, dispatch] = React.useReducer(VideoroomReader, {
```
graszka22, line 21: Ok, now why did you use `useReducer` here? `useState` was ok and simpler here :D In the example there was a counter that you could decrement and increment so I guess it was more complicated but here you just set and read room name. `useReducer` is used for complicated state.

So the context would like like this:
```
const [roomName, setRoomName] = useState()
...
<Provider value={{roomName, setRoomName}}>
...
```
And then to use it you simply do something like this:
```
const { roomName, setRoomName } = useRoomName()
```
Much simpler and obvious imho. no weird `state` and `dispatch`.

Also we can use this context for more than just a room name (for example I guess user name would be useful). So maybe instead `useRoomName` hook, we can have something like `useVideoroomState` or something.

skyman503: Okey, I changed it to be useState now, I guess this provider isn't essential right now and app could work without it, but prob. it will be useful soon so I guess it can stay.

PR #51, opened 2023-03-13
[MV-336] Add standard buttons

skyman503:
This change introduces new button component with proper styling.

### DIFF
--- a/example/src/components/buttons/StandardButton.tsx
+++ b/example/src/components/buttons/StandardButton.tsx
@@ -22,7 +22,7 @@ const StandardButtonStyles = StyleSheet.create({
   danger: {
     backgroundColor: AdditionalColors.red100,
   },
-  info: {
+  secondary: {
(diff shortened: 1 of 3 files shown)


example/src/components/buttons/StandardButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  primary: {
+    backgroundColor: BrandColors.darkBlue100,
+  },
+  danger: {
+    backgroundColor: AdditionalColors.red100,
+  },
+  info: {
```
graszka22, line 25: maybe it'd be better to name it "secondary"?

graszka22 (approved):
a minor comment, otherwise looks good 👍

example/src/components/buttons/StandardButton.tsx
skyman503: Renamed.

PR #52, opened 2023-03-14
[MV-336] Add custom text input

skyman503:
This change introduces new component that should be used, when asking user for text input.

### DIFF
--- a/example/src/components/StandardTextInput.tsx
+++ b/example/src/components/StandardTextInput.tsx
@@ -1,0 +1,89 @@
+import { AdditionalColors, BrandColors, TextColors } from '@colors';
+import { TextInputTextStyle } from '@components/Typo';
+import React, { useState } from 'react';
+import { StyleSheet, TextInput } from 'react-native';
+
+const StandardTextInputStyles = StyleSheet.create({
+  main: {
+    width: '100%',
+    height: 56,
+    borderRadius: 40,
+    borderStyle: 'solid',
+    borderWidth: 2,
+    backgroundColor: AdditionalColors.white,
+    paddingLeft: 16,
+  },
+  active: {
+    color: TextColors.darkText,
+  },
+  notActive: {
+    color: AdditionalColors.grey80,
+    borderColor: AdditionalColors.grey60,
+  },
+  offFocus: {
+    borderColor: BrandColors.darkBlue100,
+  },
+  onFocus: {
+    borderColor: BrandColors.seaBlue80,
+  },
+});
+
+type OnChangeTextType = (text: string) => void;
+
+type StandardTextInputProps = {
+  placeholder?: string;
+  value?: string;
+  editable?: boolean;
+  onChangeText?: OnChangeTextType;
+};
+
+export const StandardTextInput = ({
+  placeholder = '',
+  value = undefined,
+  editable = true,
+  onChangeText = () => {},
+}: StandardTextInputProps) => {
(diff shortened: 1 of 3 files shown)


example/src/components/StandardTextInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    return [
+      StandardTextInputStyles.main,
+      StandardTextInputStyles.notActive,
+      fontStyle,
+    ];
+  };
+
+  return (
+    <TextInput
```
graszka22, line 78: On my Android device the cursor and the selection is dark green and that doesn't look nice. Use `selectionColor` and `cursorColor` props to set it to the correct color (dark blue?)

We may need to upgrade RN version for it to look good though

example/src/components/StandardTextInput.tsx
```diff
@@ -0,0 +1,89 @@
+import { AdditionalColors, BrandColors, TextColors } from '@colors';
+import { TextInputTextStyle } from '@components/Typo';
+import React, { useState } from 'react';
+import { StyleSheet, TextInput } from 'react-native';
+
+const StandardTextInputStyles = StyleSheet.create({
```
graszka22, line 6: can't we just name it `TextInput`? Adding nothing makes it standard :p

example/src/components/StandardTextInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+type StandardTextInputProps = {
+  placeholder?: string;
+  value?: string;
+  editable?: boolean;
+  onChangeText?: OnChangeTextType;
+};
+
+export const StandardTextInput = ({
+  placeholder = '',
+  value = undefined,
```
graszka22, line 42:
```suggestion
  value,
```

graszka22 (changes requested):

example/src/components/StandardTextInput.tsx
graszka22: The handles color is broken even on the newest RN version 😅 Apparently [this](https://github.com/facebook/react-native/pull/31007) didn't fix it.
Maybe try adding it on the native side like here (in `styles.xml`): https://stackoverflow.com/questions/29795197/how-to-change-color-of-edittext-handles
If you've never done anything like this you can always ask me for help

example/src/components/StandardTextInput.tsx
skyman503: Done.

example/src/components/StandardTextInput.tsx
skyman503: Renamed styles and props.

example/src/components/StandardTextInput.tsx
skyman503: I think this should work okey now.

example/src/components/StandardTextInput.tsx
graszka22: @skyman503 What about a component?

skyman503: I though it's better not to override default names to avoid confusion later on. However, I tried doing that but it makes es-lint angry because of: `'TextInput' is already defined. eslint[@typescript-eslint/no-redeclare]` in the `StandardTextInput.tsx`. Idk whether ignoring that is a good idea. WDYT @graszka22 ?

skyman503: FIxed.

graszka22 (approved):

PR #55, opened 2023-03-16
[MV-336] Add modal

skyman503:


graszka22 (changes requested):
- how are we supposed to add buttons to this modal?
- the overlay is not semi-transparent, it covers the rest of components
- it would be nice to have this animated
- the last two points can be easily fixed using a library like https://github.com/react-native-modal/react-native-modal (it's not native, so no need to even run pod install or rebuild the app)

https://user-images.githubusercontent.com/6280457/225580305-ac27ec33-4081-4093-b7ec-3d2a7948d488.mp4

skyman503:
> 

I thought that the best way to add button to the modal is via child component, since specifying it within the props of the modal requires at least 3 new variables which would make it messy in my opinion.

I'll take care of the other two points, I genuinely thought that the transparent background was working before creating pull request. Will investigate.

graszka22:
Passing button as children is ok, I haven't noticed it 👍

skyman503:
Switched to use the library instead of the default Modal component. AFAIK animations for this component were not specified so I left the default `slideIn` one, it can be changed easily in the future if needed.

graszka22:
There is no modal exit animation (when modal is dismissed it should slide out).

example/src/components/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    marginBottom: 32,
+  },
+});
+
+type ModalProps = {
+  headline?: string;
+  body?: string;
+  visible: boolean;
+  children?: ReactNode;
+  onClose: OnPressFunctionType;
```
graszka22, line 46:
```suggestion
  onClose: () => void;
```

example/src/components/buttons/StandardButton.tsx
```diff
@@ -31,7 +31,7 @@ const StandardButtonStyles = StyleSheet.create({
 });
 
 type StandardButtonTypeName = 'primary' | 'danger' | 'secondary';
-type OnPressFunctionType = (event: GestureResponderEvent) => void;
+export type OnPressFunctionType = (event: GestureResponderEvent) => void;
```
graszka22, line 34:
```suggestion
```

example/src/components/Modal.tsx
graszka22: We don't need `GestureResponderEvent`, we just want to know when user dismissed the modal.

skyman503: Done.

example/src/components/buttons/StandardButton.tsx
skyman503: Done.

graszka22 (approved):

PR #56, opened 2023-03-16
[MV-370] Add splashscreens and icons

graszka22:
Note: if you're testing on iOS you may need to restart your device or reinstall the app


skyman503:
On my android testing device the splashscreen is not working properly. Background is white with the membrane logo, and then it's just a blank white screen until the app is loaded. Which is different than the IOS behaviour.

skyman503 (approved):

PR #58, opened 2023-03-20
[MV-340] Add Create room screen

skyman503:
This PR introduces a proper screen that should be used for creating new rooms.


example/src/App.tsx
```diff
@@ -52,13 +54,18 @@ export default function App() {
...
-          initialRouteName="Home"
+          initialRouteName="InitialScreen"
           screenOptions={{
-            headerShown: false,
+            headerBackTitle: 'Back',
           }}
         >
-          <Stack.Screen name="Home" component={Home} />
+          <Stack.Screen
+            name="CreateRoom"
```
graszka22, line 63: Changing the name of the screen from `Home` to `CreateRoom` will break deep linking (see `linking` object above). Change the linking object and verify that deeplinking still works

example/src/components/BackgroundWrapper.tsx
```diff
@@ -0,0 +1,60 @@
...
+  rightImage: {
+    width: 120,
+    height: 288,
+    alignSelf: 'flex-end',
+    resizeMode: 'stretch',
+    position: 'absolute',
+    bottom: 2,
+  },
+  content: {
+    zIndex: 3,
```
graszka22, line 51: why do we need zIndex here?

example/src/components/BackgroundWrapper.tsx
```diff
@@ -0,0 +1,60 @@
...
+    width: '100%',
+    alignSelf: 'center',
```
graszka22, line 53: I suspect that there may be too many properties here. `alignSelf: 'center'` and `position: 'absolute',` do they make sense together?

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+  View,
+  StyleSheet,
+  KeyboardAvoidingView,
+  Platform,
+  PermissionsAndroid,
+  ScrollView,
+} from 'react-native';
+import { useVideoroomState } from 'src/VideoroomContext';
+
+function isEmpty(value) {
```
graszka22, line 23: it looks like it might be useful in the future, maybe extract it to `@utils` or something
you can also add a library like `lodash` or whatever you're familiar with/is popular right now

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  useEffect(() => {
+    if (route.params?.roomName) {
+      setRoomName(route.params?.roomName);
+    }
+  }, []);
+
+  useEffect(
+    () =>
+      navigation.addListener('beforeRemove', (e) => {
```
graszka22, line 64: don't we have to remove this listener when we unmount this screen?

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+          return;
+        }
+        e.preventDefault();
+        setModalAction(e.data.action);
+        setIsModalVisible(true);
+      }),
+    [navigation, roomName, username]
+  );
+
+  const ShouldEnableCreateRoomButton = () => {
```
graszka22, line 76: why some function names start with a big letter and some with a small letter? I think the convention is to start with a small letter?

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  return (
+    <ScrollView
+      contentContainerStyle={{ flexGrow: 1 }}
+      keyboardShouldPersistTaps="handled"
+    >
+      <KeyboardAvoidingView
+        behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
+        style={{
+          flex: 1,
```
graszka22, line 149: add it to stylesheet?

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+                }}
+              />
+            </View>
+          </View>
+          <View style={{ flex: 1 }} />
+
+          <View style={styles.createRoomButton}>
+            <StandardButton
+              onPress={() => {
+                TrimTrailingSpacesForTextInputs();
```
graszka22, line 217: this is not going to work :p 
`TrimTrailingSpacesForTextInputs` changes the state (`setRoomName`). State changes are asynchronous so when we read the room name in the `connect` function we'll get the old value with spaces (for more explanation see [here](https://react.dev/reference/react/useState#ive-updated-the-state-but-logging-gives-me-the-old-value))

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+            </View>
+            <View style={styles.usernameInputLabel}>
+              <Typo variant="body-small">Your name</Typo>
+            </View>
+            <View style={styles.usernameInput}>
+              <TextInput
+                placeholder="Your name"
+                value={username}
+                onChangeText={(val) => {
+                  setUsername(val);
```
graszka22, line 207: can't we just pass `setUsername`?

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  useEffect(
+    () =>
+      navigation.addListener('beforeRemove', (e) => {
+        if (!roomName && !username) {
+          // If we don't have unsaved changes, then we don't need to do anything
+          return;
+        }
+        e.preventDefault();
+        setModalAction(e.data.action);
```
graszka22, line 70: you don't have to use state for this modal action:
- if you need to re-render components because of a change, use `useState` - for example we want to use `useState` for room name because we want the change to be visible in the ui in the text box
- if you don't need to re-render anything but just want to save a variable then use `useRef` -it's better for performance (no re-renders!) and easier to use (they're synchronous - so if you set a ref it will be set immediately like a normal variable)

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,267 @@
...
+            <View style={styles.roomInputLabel}>
+              <Typo variant="body-small">Room name</Typo>
+            </View>
+
+            <View style={styles.roomInput}>
+              <TextInput
+                placeholder="Room name"
+                value={roomName}
+                onChangeText={(val) => {
+                  setRoomName(val);
```
graszka22, line 195: can't we just pass `setRoomName`?

graszka22 (changes requested):

graszka22:
Also I don't like this behaviour: you can scroll past the blue background and see ugly white space. I think the background shouldn't be scrollable at all, just the content?

https://user-images.githubusercontent.com/6280457/226411435-0af06cda-265c-42c2-a484-2c620c547288.mp4

skyman503:
> Also I don't like this behaviour: you can scroll past the blue background and see ugly white space. I think the background shouldn't be scrollable at all, just the content?

Yes, I didn't want to make the background scrollable, but without it the `KeyboardAvoidingView` was broken, and after some tweaking I found, that's the only configuration I made it working, do you have any ideas on why is that the case?

example/src/App.tsx
skyman503: Done.

example/src/components/BackgroundWrapper.tsx
skyman503: Removed.

example/src/components/BackgroundWrapper.tsx
skyman503: Ack.

example/src/screens/CreateRoom.tsx
skyman503: using loadsh now.

example/src/screens/CreateRoom.tsx
skyman503: Added removeListiner

example/src/screens/CreateRoom.tsx
skyman503: Ack.

example/src/screens/CreateRoom.tsx
skyman503: Changed function to be async and added `await`.

example/src/screens/CreateRoom.tsx
skyman503: Done.

example/src/screens/CreateRoom.tsx
skyman503: Changed to use `useRef`.

example/src/screens/CreateRoom.tsx
skyman503: Done.

graszka22 (changes requested):

example/package.json
```diff
@@ -13,14 +13,17 @@
     "@expo/vector-icons": "^13.0.0",
     "@react-navigation/native": "^6.1.6",
     "@react-navigation/native-stack": "^6.9.12",
+    "@react-navigation/stack": "^6.3.16",
     "eslint-import-resolver-babel-module": "^5.3.2",
     "eslint-plugin-import": "^2.27.5",
     "expo": "^47.0.0",
     "expo-font": "^11.1.1",
     "expo-splash-screen": "^0.18.1",
+    "loadsh": "^0.0.4",
```
graszka22, line 22: also it's `lodash` not `loadsh`

graszka22 (changes requested):

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,262 @@
...
+              <TextInput
+                placeholder="Your name"
+                value={username}
+                onChangeText={setUsername}
+              />
+            </View>
+            <View style={styles.createRoomButton}>
+              <StandardButton
+                onPress={async () => {
+                  await trimTrailingSpacesForTextInputs();
```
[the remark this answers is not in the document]
skyman503: Added names as parameters.

graszka22 (changes requested):

example/src/screens/CreateRoom.tsx
skyman503: Done.

example/src/screens/CreateRoom.tsx
```diff
@@ -0,0 +1,257 @@
...
+            <View style={styles.usernameInput}>
+              <TextInput
+                placeholder="Your name"
+                value={username}
+                onChangeText={setUsername}
+              />
+            </View>
+            <View style={styles.createRoomButton}>
+              <StandardButton
+                onPress={() => connect(roomName.trimEnd(), username.trimEnd())}
```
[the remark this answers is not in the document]
skyman503: Good idea, thanks!

graszka22 (approved):
👏

PR #62, opened 2023-03-23
Fix animations

skyman503:
This changes fix the background in place during transitions between screens.

### DIFF
--- a/example/src/screens/CreateRoom.tsx
+++ b/example/src/screens/CreateRoom.tsx
@@ -33,7 +33,7 @@ export const CreateRoom = ({ navigation, route }: Props) => {
   const [isModalVisible, setIsModalVisible] = useState(false);
   const modalAction = useRef<GoBackAction>();
   const { roomName, setRoomName, username, setUsername } = useVideoroomState();
-  const { next } = useCardAnimation();
+  const { next, current } = useCardAnimation();
 
   useEffect(() => {
     if (route.params?.roomName) {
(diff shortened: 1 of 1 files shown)


example/src/shared/utils.ts
```diff
@@ -10,3 +10,12 @@ export const getShortUsername = (username: string) => {
     .join('')
     .toUpperCase();
 };
+
+/**
+ * Used to check whether string contains only whitespaces.
```
graszka22, line 15: No need to document that, it's in the function name and the types.
Comments like this make code less readable. In time the function might change and we'll most likely forget to update the comment.

Good comments are explaining things that are not obvious (for example "I'm doing this like that because there is an issue #1234 with RN"). What the function does should be obvious from the function name and you're doing a good job creating small, well named functions 👍 

Also I'm adding comments like this one but to the public API to automatically generate public docs. But this function not a public API.

example/src/screens/CreateRoom.tsx
```diff
@@ -90,7 +117,26 @@ export const CreateRoom = ({ navigation, route }: Props) => {
   }, []);
 
   return (
-    <BackgroundWrapper hasHeader>
+    <Animated.View
```
graszka22, line 120: I think it's a bit repetitive. Can we extract this to a separate component and reuse?

graszka22 (changes requested):

example/src/shared/utils.ts
skyman503: Remved.

example/src/screens/CreateRoom.tsx
skyman503: Added a new component.

graszka22 (approved):

PR #63, opened 2023-03-23
[MV-382] Add ability to switch between cameras using camera names

graszka22:
Also allows toggling camera in preview


skyman503 (approved):
lgtm

PR #64, opened 2023-03-23
[MV-356] Add Join room screen

skyman503:


example/src/components/TextInput.tsx
```diff
@@ -69,18 +86,39 @@ export const TextInput = ({
...
-      value={value}
-      onFocus={onFocus}
-      onBlur={offFocus}
-      editable={editable}
-      onChangeText={onChangeText}
-      // @ts-ignore
-      colorCursor={TextColors.darkText}
-      selectionColor={TextColors.additionalLightText}
-    />
+    <>
```
graszka22, line 89: Wrap it in a `<View`. I suppose it won't happen with the current designs, but if the parent container would have `flexDirection: row` the label would be to right of the text input. We wouldn't like that

example/src/screens/JoinRoom.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const { next, current } = useCardAnimation();
+
+  useEffect(() => {
+    if (route.params?.roomName) {
+      setRoomName(route.params?.roomName);
+      setIsRoomNameInputEditable(false);
+    }
+  }, []);
+
+  useEffect(() => {
```
graszka22, line 50: You're repeating code here. I propose creating a separate component like `<DiscardModal />` that handles the navigation stuff too. You can use `useNavigation` hook for accessing the navigation object outside the screen component.

example/src/screens/JoinRoom.tsx
```diff
@@ -0,0 +1,235 @@
...
+  }, []);
+
+  const openPreview = () => {
+    if (checkIfUrl(roomName)) {
+      setRoomName(decodeURI(extractRoomNameFromUrl(roomName)));
+    }
+    navigation.push('Preview', { title: 'Join meeting' });
+  };
+
+  const requestPermissionsAndOpenPreview = useCallback(async () => {
```
graszka22, line 88: This function should be extracted somewhere else, even to a separate file, you're repeating it.

example/src/shared/utils.ts
```diff
@@ -1,3 +1,7 @@
+import { isEmpty } from 'lodash';
+
+const VIDEOROOM_URL = 'https://videoroom.membrane.work/room/';
```
graszka22, line 3: we're checking this url for deeplinking too, use it there too

and better yet put it in the .env files and get it from there

example/src/shared/utils.ts
```diff
@@ -14,3 +18,20 @@ export const getShortUsername = (username: string) => {
...
+
+export const checkIfUrl = (val: string) => {
+  return val.startsWith(VIDEOROOM_URL);
+};
+
+export const extractRoomNameFromUrl = (url: string) => {
+  return url.substring(url.indexOf(VIDEOROOM_URL) + VIDEOROOM_URL.length);
+};
+
+export const shouldEnableRoomButton = (username, roomName) => {
```
graszka22, line 30: Maybe `checkIfStringContainsOnlyWhitespaces` could check if string is empty too and this function would be simpler and reduced to a single if statement? Also we could rename `checkIfStringContainsOnlyWhitespaces` to `isEmptyStringOrWhitespaces` or something like that to be more clear.
I'm not saying it's wrong putting it here, it's just weird to me putting this function in this file, because `@utils` is supposed to hold functions useful in the whole app, this function is useful only in two specific screens.

graszka22 (changes requested):

example/src/components/TextInput.tsx
skyman503: Done.

example/src/screens/JoinRoom.tsx
skyman503: Done.

example/src/shared/utils.ts
skyman503: Added env variable.

example/src/shared/utils.ts
skyman503: Removed and renamed.

graszka22 (changes requested):

example/src/screens/JoinRoom.tsx
```diff
@@ -0,0 +1,178 @@
...
+            flex: 1,
+          }}
+          enabled
+          keyboardVerticalOffset={height}
+        >
+          <DiscardModal
+            headline="Cancel joining meeting"
+            body="Are you sure you don't want to cancel joining to this meeting?"
+            onPress={() => {
+              setRoomName('');
```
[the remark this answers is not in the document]
skyman503: Moved logic to DiscardModal and removed onPress prop.

graszka22 (approved):
👏

PR #70, opened 2023-03-28
[MV-342] Add room screen

skyman503:
This changes aligns `Room` screen with the designs. `Others` / ` + `+ k people` tile with proper styling will be introduced in a separate PR.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -175,6 +175,6 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:${versions.coroutines}"
-  implementation 'com.github.jellyfish-dev:membrane-webrtc-android:4.2.0'
+  implementation 'com.github.jellyfish-dev:membrane-webrtc-android:4.2.1'
   implementation 'com.twilio:audioswitch:1.1.7'
 }
--- a/android/src/main/java/com/reactnativemembrane/MembranePackage.kt
+++ b/android/src/main/java/com/reactnativemembrane/MembranePackage.kt
@@ -1,6 +1,5 @@
 package com.reactnativemembrane
 
-import VideoPreviewViewManager
 import com.facebook.react.ReactPackage
 import com.facebook.react.bridge.NativeModule
 import com.facebook.react.bridge.ReactApplicationContext
--- a/android/src/main/java/com/reactnativemembrane/VideoPreviewView.kt
+++ b/android/src/main/java/com/reactnativemembrane/VideoPreviewView.kt
@@ -1,3 +1,5 @@
+package com.reactnativemembrane
+
 import org.membraneframework.rtc.media.LocalVideoTrack
 import org.membraneframework.rtc.ui.VideoTextureViewRenderer
 import org.webrtc.PeerConnectionFactory
@@ -5,13 +7,31 @@ import android.content.Context
 import org.membraneframework.rtc.media.VideoParameters
 import org.webrtc.EglBase
 
-class VideoPreviewView(private val context: Context) {
-  var view: VideoTextureViewRenderer = VideoTextureViewRenderer(context)
+class VideoPreviewView(private val context: Context): VideoTextureViewRenderer(context) {
   private lateinit var localVideoTrack: LocalVideoTrack
   private lateinit var eglBase: EglBase
   private lateinit var peerConnectionFactory: PeerConnectionFactory
 
-  fun init() {
+  fun switchCamera(captureDeviceId: String) {
+    localVideoTrack.switchCamera(captureDeviceId)
+  }
+
+  private fun dispose() {
+    localVideoTrack.removeRenderer(this)
+    localVideoTrack.stop()
+    localVideoTrack.rtcTrack().dispose()
+    this.release()
+    peerConnectionFactory.dispose()
+    eglBase.release()
+  }
+
+  override fun onDetachedFromWindow() {
+    super.onDetachedFromWindow()
+    dispose()
+  }
+
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
     PeerConnectionFactory.initialize(PeerConnectionFactory.InitializationOptions.builder(context).createInitializationOptions())
     eglBase = EglBase.create()
     peerConnectionFactory = PeerConnectionFactory.builder().createPeerConnectionFactory()
@@ -19,20 +39,7 @@ class VideoPreviewView(private val context: Context) {
     localVideoTrack = LocalVideoTrack.create(context, peerConnectionFactory, eglBase, VideoParameters.presetFHD169).also {
       it.start()
     }
-    view.init(eglBase.eglBaseContext, null)
-    localVideoTrack.addRenderer(view)
-  }
-
-  fun switchCamera(captureDeviceId: String) {
-    localVideoTrack.switchCamera(captureDeviceId)
-  }
-
-  fun dispose() {
-    localVideoTrack.removeRenderer(view)
-    localVideoTrack.stop()
-    localVideoTrack.rtcTrack().dispose()
-    view.release()
-    peerConnectionFactory.dispose()
-    eglBase.release()
+    init(eglBase.eglBaseContext, null)
+    localVideoTrack.addRenderer(this)
   }
 }
--- a/android/src/main/java/com/reactnativemembrane/VideoPreviewViewManager.kt
+++ b/android/src/main/java/com/reactnativemembrane/VideoPreviewViewManager.kt
@@ -1,20 +1,18 @@
+package com.reactnativemembrane
+
 import android.view.View
 import com.facebook.react.uimanager.SimpleViewManager
 import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.annotations.ReactProp
+import com.reactnativemembrane.VideoPreviewView
 import org.membraneframework.rtc.ui.VideoTextureViewRenderer
 import org.webrtc.RendererCommon
 
 class VideoPreviewViewManager : SimpleViewManager<View>() {
   override fun getName() = "VideoPreviewView"
 
-  private val viewsWrappers = HashMap<VideoTextureViewRenderer, VideoPreviewView>()
-
   override fun createViewInstance(reactContext: ThemedReactContext): View {
-    val wrapper = VideoPreviewView(reactContext)
-    wrapper.init()
-    viewsWrappers[wrapper.view] = wrapper
-    return wrapper.view
+    return VideoPreviewView(reactContext)
   }
 
   @ReactProp(name = "videoLayout")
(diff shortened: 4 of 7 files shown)


example/src/App.tsx
```diff
@@ -66,7 +66,7 @@ export default function App() {
 
   return (
     <BackgroundWrapper>
-      <SafeAreaView style={{ flex: 1 }} edges={['bottom']}>
+      <SafeAreaView style={{ flex: 1 }} edges={[]}>
```
graszka22, line 69: this SafeAreaView should be around every screen, not here

example/src/components/CallControls.tsx
```diff
@@ -39,38 +39,28 @@ export const Controls = ({ disconnect }: { disconnect: () => void }) => {
...
+      <View style={styles.iconInRow}>
         <InCallButton
-          iconName="Screen-off"
+          iconName={!isScreencastOn ? 'Screenshare' : 'Screen-off'}
           onPress={toggleScreencastAndUpdateMetadata}
         />
-      )}
+      </View>
+      <View style={styles.iconInRow}>
+        <InCallButton iconName="Menu-vertical" onPress={() => {}} />
```
graszka22, line 61: we don't need this button for now

example/src/components/RoomParticipant.tsx
```diff
@@ -0,0 +1,94 @@
...
+import React from 'react';
+import { View, StyleSheet, ViewStyle } from 'react-native';
+
+import { Icon } from './Icon';
+import { NoCameraView } from './NoCameraView';
+import { Typo } from './Typo';
+
+type RoomParticipantProps = {
+  participant: Membrane.Participant;
+  pStyle: ViewStyle[];
```
graszka22, line 13: Why `pStyle` and not just `style`? 
Also for styles like positioning etc. you can just wrap `RoomParticipant` in a styled view and avoid passing a style prop, wouldn't it be easier?
In general the component shouldn't care about its position and margins and the parent component cares about position and margins of the child component. That makes components more reusable.

example/src/components/RoomParticipant.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+export const RoomParticipant = ({
+  participant,
```
graszka22, line 17: we can destructure `participant` here, avoiding repeating `participant.something` all the time

example/src/components/RoomParticipant.tsx
```diff
@@ -0,0 +1,94 @@
...
+  pStyle,
+}: RoomParticipantProps) => {
+  return (
+    <View style={pStyle}>
+      {!participant.tracks.filter((t) => t.type === 'Video').length ||
+      !participant.tracks.find((t) => t.type === 'Video')!.metadata.active ? (
```
graszka22, line 23: I'd extract this check into a small function in this component, it would be more readable. Like `participantHasVideo()`. Or even a boolean variable. You can also get video track earlier, make an if and avoid `!` operator. Using `!` operator is a bad practice. Also you'd avoid repeating `participant.tracks.find((t) => t.type === 'Video')`

example/src/components/RoomParticipant.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <View style={styles.videoTrack}>
+          <NoCameraView
+            username={getShortUsername(participant.metadata.displayName)}
```
graszka22, line 26: wouldn't it be nicer if `NoCameraView` took username and run `getShortUsername`?

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,187 @@
...
+import React, { useCallback, useEffect } from 'react';
+import { StyleSheet, View, Pressable, Dimensions } from 'react-native';
+import { useVideoroomState } from 'src/VideoroomContext';
 
-import { Controls } from '../Controls';
-import { Settings } from '../Settings';
+import { CallControls } from '../components/CallControls';
+
+const HEADER_AND_FOOTER_SIZE = 126;
+const PADDING_BETWEEN_PARTICIPANTS = 16;
```
graszka22, line 16: shouldn't it be 8px? https://www.figma.com/file/lVPaILYT73AJIOsoXNbsXj/Membrane_Videoroom_Ready_Designs?node-id=628%3A18043&t=972uWMzRl47USrm4-0

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,187 @@
...
+    Math.ceil(participants.length / 2);
 
   const { disconnect: mbDisconnect } = Membrane.useMembraneServer();
 
   useEffect(() => {
-    if (!focusedTrack && tracks[0]) {
-      setFocusedTrackId(tracks[0].id);
-    }
-  }, [tracks, focusedTrack]);
+    const handleBeforeRemoveEvent = (e) => {
```
graszka22, line 37: I do not understant why do we need to do all this handling. Can't we just set `gestureEnabled = false` on the screen?

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,187 @@
...
-              {!focusedTrack.metadata.active && (
-                <Icon name="Cam-disabled" size={24} />
+    <BackgroundAnimation>
+      <View style={styles.container}>
+        <View style={styles.header}>
+          <View style={styles.headerTitle}>
+            <Typo variant="h4">{roomName}</Typo>
+          </View>
+          <View style={styles.headerIcon}>
+            <Pressable onPress={switchCamera}>
```
graszka22, line 88: maybe use touchable opacity to provide some feedback to the user?

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,187 @@
...
+                  <RoomParticipant
+                    key={p.id}
+                    participant={p}
+                    pStyle={getStylesForParticipants(participants)}
+                  />
+                ))}
+
+              {participants.length > MAX_NUM_OF_USERS_ON_THE_SCREEN && (
+                <View style={getStylesForParticipants(participants)}>
+                  <Typo variant="label">Others</Typo>
```
graszka22, line 125: it appears to be broken / not finished and there some new margins?
Screenshot_20230328_161719

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,187 @@
...
-    flex: 1,
-  },
-  focusedParticipantContainer: {
+  container: {
     flex: 1,
-    borderRadius: 4,
-    borderWidth: 2,
-    borderColor: '#001A72',
-    margin: 20,
+    flexDirection: 'column',
```
graszka22, line 140: it's 'column' by default

graszka22 (changes requested):

example/src/screens/Room.tsx
skyman503: I said in the PR description that I planned to submit it it a different PR since making it look/work properly won't be a small thing and this PR is already big. Wdyt?

example/src/screens/Room.tsx
skyman503: It is in the actual css, this constant is used for calculating size of the video tiles and using 8px per row wasn't enough to get tiles to fit, looking at this right now I agree that the naming is bad, I can change it though.

example/src/components/CallControls.tsx
skyman503: Ok, I'll remove it, it was part of the design so I added it.

example/src/screens/Room.tsx
graszka22: ok, do it in the separate PR 👍 
but please look into the margin issue

example/src/screens/Room.tsx
graszka22: ok, please change the naming then, now it's really misleading to me

example/src/components/RoomParticipant.tsx
skyman503: Thanks for this comment, I agree it makes sens to keep the margin styling out of the component, should be ok now.

example/src/components/RoomParticipant.tsx
skyman503: Done.

example/src/components/RoomParticipant.tsx
skyman503: Done.

example/src/screens/Room.tsx

example/src/components/RoomParticipant.tsx
skyman503: Done.

example/src/screens/Room.tsx
skyman503: Fixed margin.

example/src/screens/Room.tsx
skyman503: Ack.

example/src/screens/Room.tsx
graszka22: ok, add a comment with a link to this issue

example/src/screens/Room.tsx

example/src/components/RoomParticipant.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const RoomParticipant = ({
+  tracks,
+  metadata,
+  type,
+}: RoomParticipantProps) => {
+  const videoTrack = tracks.find((t) => t.type === 'Video');
+  const audioTrack = tracks.find((t) => t.type === 'Audio');
+
+  const participantHasVideo = () => {
+    if (videoTrack !== undefined) {
```
graszka22, line 25: Why not `if(videoTrack)`? In js values might be `undefined`, `null` and other weird values. We don't want to think about them and `if(videoTrack)` should cover them all.

example/src/components/RoomParticipant.tsx
```diff
@@ -0,0 +1,100 @@
...
+  tracks: Membrane.Track[];
+  metadata: Membrane.Metadata;
+  type: Membrane.ParticipantType;
+};
+
+export const RoomParticipant = ({
+  tracks,
+  metadata,
+  type,
+}: RoomParticipantProps) => {
```
graszka22, line 20: you can pass the whole participant and destructure it here, something like that (haven't tested it):
```
export const RoomParticipant = ({{participant: {tracks, metadata, type}}) => {
...
}
```
if I add something to the participant object later or you'll want to change `RoomParticipant` component that would be less changes.

example/src/screens/Preview.tsx
```diff
@@ -122,65 +123,61 @@ export const Preview = ({ navigation, route }: Props) => {
...
-          ) : (
-            <View style={styles.noCameraBackground}>
-              <View style={styles.noCameraContent}>
-                <Typo variant="h5" color={BrandColors.darkBlue80}>
-                  {getShortUsername(username)}
-                </Typo>
-              </View>
-            </View>
-          )}
+      <SafeAreaView style={{ flex: 1 }} edges={['bottom']}>
```
graszka22, line 126: I'm wondering if we'd like to have a `ScreenWrapper` component with `SafeAreaView` and maybe something else in the future. If we'd have to change the safe area we'd need to change in every screen again.

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,192 @@
...
+import { StyleSheet, View, Dimensions } from 'react-native';
+import { TouchableOpacity } from 'react-native-gesture-handler';
+import { useVideoroomState } from 'src/VideoroomContext';
 
-import { Controls } from '../Controls';
-import { Settings } from '../Settings';
+import { CallControls } from '../components/CallControls';
+
+const HEADER_AND_FOOTER_SIZE = 126;
+const OFFSETPER_ROW = 16;
```
graszka22, line 17:
```suggestion
const OFFSET_PER_ROW = 16;
```

example/src/screens/Room.tsx
```diff
@@ -1,181 +1,192 @@
...
-                <Icon name="Cam-disabled" size={24} />
-              )}
-            </View>
-            <Pressable
-              style={styles.settingsButton}
-              onPress={() => setAreSettingsOpen((o) => !o)}
-            >
-              <Icon name="Settings" size={48} />
-            </Pressable>
+    <BackgroundAnimation>
```
graszka22, line 82: no safe area here?

graszka22 (changes requested):

example/src/components/RoomParticipant.tsx
skyman503: Fixed.

example/src/screens/Room.tsx
skyman503: Added.

example/src/components/RoomParticipant.tsx
skyman503: Done.

graszka22 (approved):
👏

PR #71, opened 2023-03-28
Stop camera in preview when going to another screen

graszka22:
It fixes the following bug on Android (and a similar one on iOS):
- go from preview screen to room screen
- go back
- preview is broken

Also helps to conserve resources.
I also refactored a little bit by the way.


skyman503 (approved):
lgtm thanks!

PR #72, opened 2023-03-29
[MV-384] Improve animation when moving between preview and room screen

graszka22:
Let's merge https://github.com/jellyfish-dev/react-native-membrane-webrtc/pull/70 first and then test with the new room screen.


skyman503:
Could you rebase with HEAD?

PR #74, opened 2023-03-29
[MV-342] Add other participants tile

skyman503:


example/src/components/OtherParticipants.tsx
```diff
@@ -0,0 +1,99 @@
+import { BrandColors } from '@colors';
+import * as Membrane from '@jellyfish-dev/react-native-membrane-webrtc';
+import { getShortUsername } from '@utils';
+import React from 'react';
+import { View, StyleSheet } from 'react-native';
+
+import { Typo } from './Typo';
+
+type OtherParticipantsProp = {
+  p1: Membrane.Participant | undefined;
```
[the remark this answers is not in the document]
skyman503: Ack.

example/src/screens/Room.tsx
```diff
@@ -119,15 +120,24 @@ export const Room = ({ navigation }: Props) => {
...
                   ))}
 
                 {participants.length > MAX_NUM_OF_USERS_ON_THE_SCREEN && (
                   <View style={getStylesForParticipants(participants)}>
-                    <Typo variant="label">Others</Typo>
+                    <OtherParticipants
+                      p1={participants[participants.length - 1]}
+                      p2={participants[participants.length - 2]}
+                      numOfOtherParticipants={
+                        participants.length - MAX_NUM_OF_USERS_ON_THE_SCREEN
```
graszka22, line 138: Ok, one more bug: the number of other participants is invalid. For 10 participants there is no text (7 participants on screen, 2 circles so the text should say 1 other participant) and for more there is 1 participant missing.

graszka22 (changes requested):

example/src/screens/Room.tsx
skyman503: Good catch, I forgot that when others tab appears there's one less space for a regular participant. Should be fixed now.

graszka22 (approved):

PR #75, opened 2023-03-29
Add Leave room screen

graszka22:


example/src/components/Typo.tsx
```diff
@@ -172,12 +178,14 @@ type TypoProps = {
...
   color?: string;
   children: ReactNode;
-};
+} & TextProps;
 
 export const Typo = ({
   variant = 'body-big',
   color = TextColors.darkText,
   children,
+  style,
```
skyman503, line 187: Are you sure that adding custom styling to `Typo` component is a good practice? I would have assumed that we should keep it clean from custom behaviour since it was predefined in the designs and all of the additional styling should be done around it using `<View>` components.

example/src/screens/LeaveRoomScreen.tsx
```diff
@@ -0,0 +1,78 @@
...
+    } catch (err) {
+      console.warn(err);
+    }
+  }, [connectAndJoinRoom]);
+
+  return (
+    <View style={styles.content}>
+      <Image
+        style={styles.logo}
+        source={require('../../assets/images/Logo.png')}
```
skyman503, line 31: Mby it's time for an `@assets` alias, wdyt?

example/src/components/Typo.tsx
skyman503: Sg.

PR #76, opened 2023-03-30
[MV-346] Add pin functionality

skyman503:


example/src/components/OtherParticipants.tsx
```diff
@@ -47,7 +47,10 @@ export const OtherParticipants = ({
       )}
       {numOfOtherParticipants > 2 ? (
         <View style={styles.textRow}>
-          <Typo variant="label">+ {numOfOtherParticipants - 2} others</Typo>
+          <Typo variant="label">
+            + {numOfOtherParticipants - 2} other
+            {numOfOtherParticipants - 2 > 1 ? 's' : null}
```
graszka22, line 52: FYI for such things we have libraries like this one: https://www.i18next.com/translation-function/plurals
but we are not going to add a whole i18n library just for avoiding an if :D so it can stay here

example/src/components/RoomParticipant.tsx
```diff
@@ -24,40 +35,75 @@ export const RoomParticipant = ({
...
+  const getTextForPinButton = () => {
+    return focused ? 'Unpin user' : titleSmall ? 'Pin' : 'Pin user';
+  };
+
+  const onPinButton = () => {
+    if (focused) {
+      onPinButtonPressed(null);
+      return;
+    }
+    onPinButtonPressed(videoTrack!.id);
```
graszka22, line 47: what if I click on a user without video track (it is possible and I can reproduce it by joining with camera off)? 
that's why using `!` is a bad practice :D

example/src/components/RoomParticipant.tsx
```diff
@@ -24,40 +35,75 @@ export const RoomParticipant = ({
...
+  };
+
+  const triggerShowingPinButton = async () => {
+    if (pinButtonHiddden) {
+      return;
+    }
+
+    setShowPinButton(true);
```
graszka22, line 55: can we have this animated? a simple animation, just opacity transition

example/src/components/RoomParticipant.tsx
skyman503: Changed to use participant id instead. Videotrack id was way more convinient to use if you take screencast into consideration but there's a lot of edge cases that needed to be accounted for, so I think participant id will be the better choice.

example/src/components/RoomParticipant.tsx
skyman503: Added FadeIn animation.

example/src/components/RoomParticipant.tsx
```diff
@@ -24,40 +35,75 @@ export const RoomParticipant = ({
...
+    await delay(2000);
```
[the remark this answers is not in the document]
skyman503: I couldn't reproduce this bug myself, however I think it should be gone now. I went with a more convenient(I think) approach, where if the button is visible pressing multiple times is ignored and thous there's only one promise running at any given time. Please check whether it works for you now.

example/src/components/RoomParticipant.tsx
graszka22: You should also add fade out animation when button is disappearing

example/src/components/RoomParticipant.tsx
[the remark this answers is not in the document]
graszka22: it appears to be fixed

example/src/components/buttons/PinButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { TouchableOpacity } from 'react-native-gesture-handler';
+
+type PinButtonProp = {
+  onPress: () => void;
+  children: ReactNode;
+};
+
+export const PinButton = ({ children, onPress }: PinButtonProp) => {
+  return (
+    <TouchableOpacity onPress={onPress} style={styles.container}>
```
graszka22, line 15: when I click pin button now the touchable opacity is not rounded:

https://user-images.githubusercontent.com/6280457/229092377-efec624d-e375-4a38-8977-543f8c0c48ae.mp4

example/src/screens/Room.tsx
```diff
@@ -94,50 +106,78 @@ export const Room = ({ navigation }: Props) => {
...
-                  </View>
-                )}
+                  )}
+                </View>
               </View>
-            </View>
+            )}
           </View>
-          <CallControls />
+          <CallControls bottomOffset={useSafeAreaInsets().bottom} />
```
graszka22, line 180: please use hooks before returning components
we're often rendering things conditionally and react is going to crash if you start/stop calling hook within lifetime of a component

example/src/components/CallControls.tsx
```diff
@@ -36,7 +40,7 @@ export const CallControls = () => {
   }, []);
 
   return (
-    <View style={styles.iconsContainer}>
+    <View style={[styles.iconsContainer, { marginBottom: 34 - bottomOffset }]}>
```
graszka22, line 43: please add a constant for 34

graszka22 (changes requested):

example/src/components/RoomParticipant.tsx
graszka22: But as you said it will break for screencast, one user can have two video tracks: video and screencast.

skyman503: Yes, but I was planning to fix that with the screencast PR

example/src/components/RoomParticipant.tsx
skyman503: I tried adding onExit={FadeOut} but that breaks other styles in the call, since the focused participant is shown in the middle of the animations, I can investigate more though .

example/src/components/buttons/PinButton.tsx
skyman503: Which emulator have you used? I don't see such a behaviour on mine.

graszka22: I used my real android device

example/src/screens/Room.tsx
skyman503: Done.

example/src/components/RoomParticipant.tsx
skyman503: My new idea is to keep it with participantId and then in the screencast PR add additional parameter like `videoTrackType= 'camera' | 'screenshare'` and that's how I'will distinguish the two.

graszka22: sounds good

example/src/components/CallControls.tsx
skyman503: Done.

example/src/components/buttons/PinButton.tsx
skyman503: Should be fixed now. But please double check.

example/src/components/RoomParticipant.tsx
skyman503: Fixed.

example/src/components/RoomParticipant.tsx
```diff
@@ -24,40 +42,93 @@ export const RoomParticipant = ({
...
+              color={BrandColors.darkBlue100}
+            />
+          </View>
+        )}
+      </Pressable>
+      {showPinButton ? (
+        <Animated.View style={[styles.pinButton, opacityStyle]}>
+          <View style={styles.pinButtonWrapper}>
+            <PinButton onPress={onPinButton}>
+              {getTextForPinButton()}{' '}
```
graszka22, line 127:
```suggestion
              {getTextForPinButton()}
```

example/src/components/RoomParticipant.tsx
```diff
@@ -24,40 +42,93 @@ export const RoomParticipant = ({
...
+  const triggerShowingPinButton = async () => {
+    if (pinButtonHiddden || isPinButtonShown.current) {
+      return;
+    }
+
+    buttonOpacity.value = withTiming(1, { duration: 300 });
+    isPinButtonShown.current = true;
+    setShowPinButton(true);
+    await delay(1700);
+    buttonOpacity.value = withTiming(0, { duration: 300 });
```
graszka22, line 72: the `withTiming` function has a callback argument https://docs.swmansion.com/react-native-reanimated/docs/api/animations/withTiming#callback-functionoptional

and you have also `withDelay` and `withSequence` functions but I'm not sure if they'll be useful here
(maybe you can make a sequence of timing, delay and timing animations)

graszka22 (changes requested):

example/src/screens/Room.tsx
```diff
@@ -94,50 +107,78 @@ export const Room = ({ navigation }: Props) => {
...
-                  </View>
-                )}
+                  )}
+                </View>
               </View>
-            </View>
+            )}
           </View>
-          <CallControls />
+          <CallControls bottomOffset={bottomOffset} />
```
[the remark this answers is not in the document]
skyman503: Moved to CallControls.

example/src/components/RoomParticipant.tsx
skyman503: Done.

example/src/components/RoomParticipant.tsx
skyman503: Added proper sequence with timings.

graszka22 (approved):
👏

PR #77, opened 2023-03-30
Navigation fixes

graszka22:
Fixes:
- screens jumping because of the headers
- rejoin button in Leave screen
- clears username and room name when navigating back to main screen
- little refactor: extracted navigation to separate file

### DIFF
--- a/example/src/Navigation.tsx
+++ b/example/src/Navigation.tsx
@@ -50,7 +50,7 @@ export const Navigation = () => {
           cardStyle: { backgroundColor: 'transparent' },
         }}
       >
-        {videoroomState === 'InMeeting' ? (
+        {videoroomState === 'InMeeting' || videoroomState === 'AfterMeeting' ? (
           <>
             <Stack.Screen
               name="Room"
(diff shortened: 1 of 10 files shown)


example/src/Navigation.tsx
```diff
@@ -0,0 +1,105 @@
...
+                headerShown: false,
+              }}
+            />
+            <Stack.Screen
+              name="CreateRoom"
+              component={CreateRoom}
+              options={{
+                title: 'New meeting',
+                headerStyle: {
+                  backgroundColor: '#ff0000',
```
skyman503, line 88: Is it needed?

skyman503:
https://user-images.githubusercontent.com/46264021/229067917-19540d45-2b91-4cab-994d-c0471f444fbb.mp4

I'm not sure whether I like the animation, for me it feels kinda weird that first we go to the right and when we want to go back into the call there is also animation of going right, where I think left would be more reasonable. Wdyt?

skyman503 (approved):
lgtm, thanks!

PR #81, opened 2023-04-04
[MV-396] Add Sentry logging service

graszka22:


skyman503 (approved):
Looks pretty good to me. I never heard of this sentry service before but looking at the generated plots it seems fine to me.

PR #83, opened 2023-04-04
[MV-391] Add screencast support

skyman503:


graszka22 (changes requested):
In general the code in the Room screen is too complicated and too nested. 
Consider doing it like this:
Move the participants part to another component.

Then calculate `participantsWithTracks` like you're doing it now and keep the focused participant+track in state.
Now you have two cases:
1. If there is a focused track then return the rendered focused track+non-focused participants (you can have multiple return statements in one component!)
2. Else return the participants in a grid.

And extract the logic of getting the tracks etc from the views to functions. That should take care of most of the nesting.

About my Android device bug: the focused participant takes the whole width of the screen and the height according to aspect ratio. Then the not focused participants take the available height and they have aspect ration set to 1 so they take the same width. Probably my device is taller than yours. RN reports window size 411x936px, my model is `xperia 10 III`. 
In my opinion the non focused participants should have 128x128px and the the focused participant takes the rest of available space. I'd fix that but right now the code is too complicated for me :D

skyman503:
> In general the code in the Room screen is too complicated and too nested. Consider doing it like this: Move the participants part to another component.
> 
> Then calculate `participantsWithTracks` like you're doing it now and keep the focused participant+track in state. Now you have two cases:
> 
> 1. If there is a focused track then return the rendered focused track+non-focused participants (you can have multiple return statements in one component!)
> 2. Else return the participants in a grid.
> 
> And extract the logic of getting the tracks etc from the views to functions. That should take care of most of the nesting.
> 
> Also it would be nice if the RoomParticipant just took the whole available space and didn't care about aspect ratios etc.
> 
> About my Android device bug: the focused participant takes the whole width of the screen and the height according to aspect ratio. Then the not focused participants take the available height and they have aspect ration set to 1 so they take the same width. Probably my device is taller than yours. RN reports window size 411x936px, my model is `xperia 10 III`. In my opinion the non focused participants should have 128x128px and the the focused participant takes the rest of available space. I'd fix that but right now the code is too complicated for me :D

I did a fair bit of refactoring and in my opinion `Room.tsx` is looking clean now. I don't like the idea of roomParticipant taking the whole available space, I'll write more on why in response to one of the other comments.

example/src/components/RoomParticipant.tsx
```diff
@@ -171,6 +190,13 @@ const styles = StyleSheet.create({
     aspectRatio: 1,
     alignSelf: 'center',
   },
+  videoTrackScreencast: {
+    flex: 1,
+    aspectRatio: 1,
```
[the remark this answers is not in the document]
skyman503: The focused prop will have to stay either way since it is used to determine not only the style but also pin button behaviour and moving it to the parent would make it less readable in my opinion. Also managing aspect ratio in the parent isn't as simple as it may seem, since I want to keep the displayName / muted / pin icons in fixed positions whatever the aspect ratio is, and managing thet from the parent would affect that. That's also why I don't like the idea of room participant taking the entire space as it would depend on the parent aspect ratio.

example/src/components/RoomParticipant.tsx
```diff
@@ -18,23 +18,31 @@ import { PinButton } from './buttons/PinButton';
 
 type RoomParticipantProps = {
   participant: Membrane.Participant;
+  trackIdx?: number;
   onPinButtonPressed?: (string) => void;
   focused?: boolean;
   pinButtonHiddden?: boolean;
   tileSmall?: boolean;
+  isScreencast?: boolean;
```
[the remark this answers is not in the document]
skyman503: Removed.

example/src/components/StopScrencastingWithFocus.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { View, StyleSheet } from 'react-native';
+
+import { Typo } from './Typo';
+import { StandardButton } from './buttons/StandardButton';
+
+export const StopScreencastingWithFocus = () => {
+  const { toggleScreencast } = Membrane.useScreencast();
+
+  const toggleScreencastAndUpdateMetadata = () => {
+    toggleScreencast({
```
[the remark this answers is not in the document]
skyman503: Sorry, done.

example/src/components/StopScrencastingWithFocus.tsx
```diff
@@ -0,0 +1,66 @@
...
+        type: 'screensharing',
+        active: 'true',
+      },
+      quality: Membrane.ScreencastQuality.FHD30,
+    });
+  };
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.content}>
```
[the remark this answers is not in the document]
skyman503: Merged.

example/src/screens/Room.tsx
```diff
@@ -109,63 +149,87 @@ export const Room = () => {
...
+                        .filter(
+                          (p) =>
+                            (p.participant.id === focusedParticipant?.id &&
+                              p.idx !== focusedParticipantData?.trackIdx) ||
+                            p.participant.id !== focusedParticipant?.id
+                        )
+                        .map((p) => ({
+                          participant: p.participant,
+                          trackIdx: p.idx,
+                        }))}
```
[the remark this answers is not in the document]

example/src/screens/Room.tsx
```diff
@@ -61,6 +81,26 @@ export const Room = () => {
   const smallScreenVideoWidth =
     (height - HEADER_AND_FOOTER_SIZE - OFFSET_PER_ROW * (rowNum + 2)) / rowNum;
 
+  const findScreensharingTrack = (track: Membrane.Track) => {
```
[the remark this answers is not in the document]

example/src/screens/Room.tsx
```diff
@@ -37,11 +42,26 @@ export const Room = () => {
   }, []);
 
   const focusedParticipant = participants.find(
-    (p) => p.id === focusedParticipantId
+    (p) => p.id === focusedParticipantData?.participantId
   );
 
+  const participantsWithTracks = participants
+    .map((p) => {
+      const videoTracks = p.tracks;
```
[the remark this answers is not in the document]

example/src/screens/Room.tsx
```diff
@@ -37,11 +42,26 @@ export const Room = () => {
...
+      if (videoTracks.some((t) => t.type === Membrane.TrackType.Video)) {
+        return p.tracks
+          .filter((t) => t.metadata.type !== 'audio')
+          .map((_t, idx) => ({
+            participant: p,
+            idx,
```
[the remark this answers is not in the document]
skyman503: Changed to use trackID.

example/src/components/RoomParticipant.tsx
```diff
@@ -18,23 +18,32 @@ import { PinButton } from './buttons/PinButton';
...
   onPinButtonPressed?: (string) => void;
   focused?: boolean;
   pinButtonHiddden?: boolean;
   tileSmall?: boolean;
 };
 
 export const RoomParticipant = ({
-  participant: { metadata, tracks, type, id },
+  participant,
+  trackId = '',
```
graszka22, line 30: why is it empty string by default? empty track id doesn't make sense, use nulls if there is no video track

example/src/components/RoomParticipant.tsx
```diff
@@ -18,23 +18,32 @@ import { PinButton } from './buttons/PinButton';
...
+
   const [showPinButton, setShowPinButton] = useState(false);
   const isPinButtonShown = useRef(false);
-  const videoTrack = tracks.find((t) => t.type === 'Video');
-  const audioTrack = tracks.find((t) => t.type === 'Audio');
+
+  const videoTrack = tracks.find((t) => t.id === trackId);
+  const videoTrackType = videoTrack?.metadata.type;
+  const audioTrack =
+    !videoTrack || videoTrackType === 'camera'
```
graszka22, line 44: It's not clear what are you doing here: if there is no video track or the videotrack is from camera then search for audio track? Why? You already have a check if the track is screenshared

example/src/screens/Room.tsx
```diff
@@ -106,74 +113,25 @@ export const Room = () => {
           <View style={{ flex: 1 }}>
             {shouldShowParticipants && (
               <>
```
graszka22, line 115: We no longer need the outer fragment (`<>`)

graszka22 (changes requested):
Only a couple of simple code style fixes this time 👍

example/src/components/FocusedParticipant.tsx
```diff
@@ -0,0 +1,61 @@
...
+  focusedParticipant,
+  onPress,
+}: FocusedParticipantProp) => {
+  const focusedTrack = focusedParticipant.participant.tracks.find(
+    (t) => t.id === focusedParticipant.trackId
+  );
+
+  return (
+    <View style={styles.focusedParticipantContainer}>
+      <View style={styles.focusedParticipant}>
```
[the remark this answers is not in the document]
skyman503: Simple copy pasting won't work here, so I think it would be best to leave it as it is for now especially since MV-395 will be reworking it's styling anyway.

example/src/components/FocusedParticipant.tsx
```diff
@@ -0,0 +1,61 @@
...
+        {focusedParticipant.participant.type ===
```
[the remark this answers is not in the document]
skyman503: Done.

example/src/components/Participants.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+type ParticipantsProp = {
+  participants: Participant[];
+  onPress: (string) => void;
+};
+
+export const Participants = ({ participants, onPress }: ParticipantsProp) => {
+  const { width, height } = Dimensions.get('window');
+
+  const rowNum = Math.min(
```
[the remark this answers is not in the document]
skyman503: Done.

example/src/components/RoomParticipant.tsx
skyman503: Changed to null.

example/src/components/RoomParticipant.tsx
skyman503: I never realised I could just that screenshared check from before, good point. Should be simpler now.

example/src/screens/Room.tsx
skyman503: I don't think so. I tried removing that and it gives an error, which the editor constantly recommends to fix by wrapping in empty jsx tag.

example/src/components/RoomParticipant.tsx
```diff
@@ -18,23 +18,32 @@ import { PinButton } from './buttons/PinButton';
...
+
   const [showPinButton, setShowPinButton] = useState(false);
   const isPinButtonShown = useRef(false);
-  const videoTrack = tracks.find((t) => t.type === 'Video');
-  const audioTrack = tracks.find((t) => t.type === 'Audio');
+
+  const videoTrack = trackId ? tracks.find((t) => t.id === trackId) : null;
+  const videoTrackType = videoTrack?.metadata.type;
+  const audioTrack =
+    videoTrackType !== 'screensharing'
```
graszka22, line 44: I still think this check is unecessary (though it's simpler). You literally have the same one lower in the code (`videoTrackType !== 'screensharing' &&`)

graszka22 (changes requested):

example/src/screens/Room.tsx

skyman503:
I think useEffect in the `Room` screen solves the no videoTrack update bug we talked about, it should be all working now.

example/src/screens/Room.tsx
```diff
@@ -57,33 +58,53 @@ export const Room = () => {
...
+
+  useEffect(() => {
+    const curretStateOfFocusedParticipant = participants.find(
+      (p) => p.id === focusedParticipantData?.participant.id
+    );
+    if (
+      focusedParticipantData?.participant !== curretStateOfFocusedParticipant
+    ) {
+      setFocusedParticipantData({
+        participant: curretStateOfFocusedParticipant!,
```
graszka22, line 98: what if pinned participant exits the call? I've got a crash

graszka22 (changes requested):

example/src/screens/Room.tsx
skyman503: Good catch, thanks. I haven't realised that the participant that is being focused could disappear.

graszka22: and get rid of this ! now, it's not necessary
and it caused the crash, I'm considering adding a lint rule to prohibit using ! :D

skyman503: Removed.

graszka22 (approved):
ok now it looks pretty good 👏 
Maybe putting the whole focused participant instead of just id wasn't a good idea after all - now I see we now have kind of double state but we'll think that out and refactor this later. Sorry for that 😅

PR #84, opened 2023-04-06
[MV-395] Fix UI bugs on small devices

skyman503:
This PR aims to fix the following list of UI bugs that happened on small devices:

- Buttons on initial screen are to big
- Discard changes modal does not fit on the screen
- Header font size is to big on Create/Join Room screens
- Keyboard avoiding view does not work on Create/Join Room screens
- Preview screen is not scrollable and does not fit
- Not focused participants are too big


graszka22 (approved):
LGTM, good job 👏

PR #85, opened 2023-04-06
Fix crash on iOS when killing backend in the middle of the call

graszka22:
To reproduce:
- enter the call on ios
- kill the backend
- ios will crash


skyman503 (approved):
Never noticed that, thanks for the catch

PR #86, opened 2023-04-07
Add Jest and unit test workflow

skyman503:


example/__tests__/videoroomContext.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const membraneWebRTC = require('../../src/index');
+membraneWebRTC.useScreencast = () => {
+  return { isScreencastOn: false, toggleScreencast: () => {} };
+};
+membraneWebRTC.useMembraneServer = () => {
+  return {
+    connect: async (): Promise<void> => {
+      return new Promise<void>((resolve) => {
+        resolve();
```
graszka22, line 33: can you put this in the mocks directory? also it's the same as joinRoom, so maybe de-duplicate this

example/src/VideoroomContext.tsx
```diff
@@ -172,7 +172,7 @@ const VideoroomContextProvider = (props) => {
 function useVideoroomState() {
   const context = React.useContext(VideoroomContext);
   if (context === undefined) {
-    throw new Error('useRoomName must be used within a VideoroomContext');
+    throw new Error('context values must be used within a VideoroomContext');
```
graszka22, line 175:
```suggestion
    throw new Error('useVideoroomState must be used within a VideoroomContext');
```

src/__mocks__/index.tsx
```diff
@@ -0,0 +1,32 @@
+const NOOP = () => {
+  // noop
+};
+
+export enum VideoQuality {
```
graszka22, line 5: is there some way to import this from the real `index.tsx`? we're going to forget to change this here

src/__mocks__/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  QVGA_43 = 'QVGA43',
+  VGA_43 = 'VGA43',
+  QHD_43 = 'QHD43',
+  HD_43 = 'HD43',
+  FHD_43 = 'FHD43',
+}
+
+export const useMembraneServer = NOOP;
+export const useRoomParticipants = NOOP;
+export const useCameraState = NOOP;
```
graszka22, line 20: It would be better if the hooks returned objects with mocked data / functions. Users don't just call the hooks, they also read something from them

graszka22 (changes requested):

example/__tests__/videoroomContext.test.ts
skyman503: Done.

src/__mocks__/index.tsx
skyman503: Added import and export for all of the enums

src/__mocks__/index.tsx
skyman503: Done.

graszka22 (approved):
Looks good 👏

PR #87, opened 2023-04-07
[MV-364] Fix screencast crashes on Android

graszka22:
To reproduce crashes turn the screencast on / off quickly.


skyman503 (approved):
lgtm, thanks

PR #88, opened 2023-04-12
[MV-404] A temporary fix for broken ios screencast

graszka22:
Please test it on some iOS device. This bug is quite non deterministic.
Make sure that the other side is receiving both screencast and video from the camera.


skyman503:
I tried it a few times on iPhone 14 IOS 16.3 and for me it still does not always work. Additionally' I've found another bug related to screencast:  

1. Join the room
2. Start screencasting
3. Disconnect without manually stoping screencast
4. Screencast is broken as ios does not know that we've left the call and notification about screencasting is still present, and another request for screencasting will fail(eg. rejoining the room and triggering screencast).

but it can be fixed later since it's not related to simulcast.

skyman503 (changes requested):
PTAL at the ios 16.3

PR #91, opened 2023-04-13
[MV-400] Add more unit tests

skyman503:


example/__tests__/videoroomContext.test.ts
```diff
@@ -108,4 +129,68 @@ describe('Videoroom context', () => {
...
+      result.current.setRoomName('test room     ');
+      result.current.setUsername('test user   ');
+    });
+    await act(async () => {
+      await result.current.connectAndJoinRoom();
+    });
+
+    expect(result.current.roomName).toBe('test room');
+    expect(result.current.username).toBe('test user');
+    expect(setExtra).toBeCalledTimes(2);
```
graszka22, line 148: can we check if we set correct values here?

example/__tests__/videoroomContext.test.ts
```diff
@@ -108,4 +129,68 @@ describe('Videoroom context', () => {
...
+      wrapper: VideoroomContextProvider,
+    });
+
+    await act(async () => {
+      await result.current.connectAndJoinRoom();
+      await result.current.disconnect();
+    });
+
+    expect(disconnectCallback).toBeCalledTimes(1);
+    expect(result.current.videoroomState).toBe('AfterMeeting');
```
graszka22, line 165: I'd add a test for rejoining, it's a bit tricky. Also a test for going back to the main screen (the name and room name are reset).

graszka22 (commented):
It's ok, two suggestions

example/__tests__/videoroomContext.test.ts
skyman503: Done.

example/__tests__/videoroomContext.test.ts
skyman503: Added rejoining scenario. 
There was already a test checking for going back to the main screen behaviour. I updated it to call `disconnect` before leaving. I think that's enough given that those are supposed to be unit tests only.

graszka22 (approved):

PR #93, opened 2023-04-17
[MV-405] Fix deeplinking when app is in background

graszka22:
The bug can be reproduced by trying to click a link when app was in Room or Leave screen. The JoinRoom screen was not mounted at all so there was no way to navigate to it.
Now all screens are in the same stack, not conditionally rendered with this fix: https://github.com/react-navigation/react-navigation/pull/11315

Pro tip for fast testing on android: `adb shell am start -W -a android.intent.action.VIEW -d "https://videoroom.membrane.work/room/Membrane%20Frontend%20Stand-up" com.membrane.reactnativemembrane` simulates a deeplink click
on iOS there is a similar command


skyman503 (commented):
The changes look ok to me and they works as you described it. However I'm not sure about the behaviour when user is currently in call. In my opinion in that case we should first display to the user some popup/modal informing them that they're about to leave a call before force moving them to the join room screen. Wdyt?

graszka22:
@skyman503 good idea, I've added the modal 👍

skyman503 (approved):
looks good now, thanks!

PR #97, opened 2023-04-25
[MV-409] Add Dev mode

skyman503:
Long pressing on initial screen enables developer mode for the user

### DIFF
--- a/example/src/VideoroomContext.tsx
+++ b/example/src/VideoroomContext.tsx
@@ -37,6 +37,7 @@ const VideoroomContext = React.createContext<
       setIsDevMode: (isDevMode: boolean) => void;
       devServerUrl: string;
       setDevServerUrl: (devServerUrl: string) => void;
+      setSavedIsDevMode: (updatedIsDevMode: boolean) => Promise<void>;
       connectAndJoinRoom: () => Promise<void>;
       disconnect: () => Promise<void>;
       videoroomState: VideoroomState;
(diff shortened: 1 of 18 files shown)


example/src/screens/InitialScreen.tsx
```diff
@@ -3,19 +3,24 @@ import { Typo } from '@components/Typo';
...
   return (
     <BackgroundAnimation>
       <SafeAreaView style={{ flex: 1 }} edges={['bottom']}>
         <View style={styles.content}>
           <View style={styles.header}>
-            <Image
-              style={styles.logo}
-              source={require('@assets/images/Logo.png')}
-            />
+            <Pressable onLongPress={() => setIsDevMode(!isDevMode)}>
```
graszka22, line 18: can we have a confirmation modal or something? the way it is right now you might not notice that you're in dev mode

example/src/VideoroomContext.tsx
```diff
@@ -48,6 +52,8 @@ const VideoroomContextProvider = (props) => {
   const [currentCamera, setCurrentCamera] = useState<CaptureDevice | null>(
     null
   );
+  const [isDevMode, setIsDevMode] = useState(false);
```
graszka22, line 55: can we save it, so that we don't need to turn it on again after refresh?

graszka22 (changes requested):

example/src/VideoroomContext.tsx
skyman503: I can use `AsyncStorage` lib for that, I wasn't sure whether it's worth the extra effort. If you prefer I can add it

graszka22: I think adding async storage lib not a big deal, we're probably going to need it in the future anyway. And it will be useful, I forgot to turn on the dev mode a couple of times already.

example/src/screens/InitialScreen.tsx
skyman503: Done.

example/src/screens/InitialScreen.tsx
```diff
@@ -1,21 +1,42 @@
...
+
+  const toggleDevMode = async () => {
+    const updatedIsDevMode = !isDevMode;
+    Alert.alert(
+      'Dev mode is now '.concat(updatedIsDevMode ? 'enabled' : 'disabled')
+    );
+    setIsDevMode(updatedIsDevMode);
+    try {
+      const jsonValue = JSON.stringify(updatedIsDevMode);
+      await AsyncStorage.setItem('isDevMode', jsonValue);
```
graszka22, line 23: Put it in setIdDevMode function in the context.The context is already managing reading the value, so it should also manage writing the value so it's all in one place.

graszka22 (changes requested):

example/src/screens/InitialScreen.tsx
skyman503: I did some magic there, since async store is... async it cannot be simply used in the useEffect. Wdyt about my current approach?

skyman503: Sg!

graszka22 (approved):
👏

PR #98, opened 2023-04-27
[MV-418] Fix buttons overlap on Android

graszka22:


example/src/components/CallControls.tsx
```diff
@@ -29,12 +26,7 @@ export const CallControls = () => {
   }, [disconnect]);
 
   return (
-    <View
-      style={[
-        styles.iconsContainer,
-        { marginBottom: BOTTOM_MARGIN - bottomOffset },
-      ]}
-    >
+    <View style={[styles.iconsContainer, { marginBottom: BOTTOM_MARGIN }]}>
```
skyman503, line 29: What about moving `marginBottom` to the `styles.iconsContainer` and getting rid of `BOTTOM_MARGIN` variable

skyman503 (changes requested):

example/src/components/CallControls.tsx
graszka22: Good catch!

skyman503 (approved):

PR #99, opened 2023-04-28
[MV-433] Fix Android app not always switching to bluetooth headphones

graszka22:
`BLUETOOTH_CONNECT` permission is no longer necessary.
Please test especially the scenario on the first run of the application. The app should automatically switch to bluetooth headphones if they're connected. Otherwise it should be switched to speakersphone.


graszka22:
I am able to gradle sync and build the app normally. Could you show me a screenshot of the error and the build file?

skyman503:
Works fine now, I did some gradle debuging on my own and ofc some random cache was causing a problem for me. LGTM thanks :D

skyman503 (approved):

PR #102, opened 2023-05-10
[MV-438] Fix long room name overlapping with flip button in Room screen

skyman503:


example/src/screens/Room.tsx
```diff
@@ -197,6 +199,7 @@ const styles = StyleSheet.create({
   },
   headerTitle: {
     marginLeft: 16,
+    maxWidth: '80%',
```
graszka22, line 202: Why 80%? Shouldn't it take just any available space? I can fit a few more letters easily here:
Screenshot_20230510_145221

skyman503: changed to flex

graszka22 (approved):

PR #104, opened 2023-05-11
[RTC-239] Refactor for Jellyfish client

graszka22:
- create and manage socket on the JS side
- in Jellyfish we'll create a different `connect` function with another socket and use that socket for backend communication. We'll use the `receiveMediaEvent` function and `onSendMediaEvent` callback (or event in case of RN) to communicate with WebRTC part
- join function now takes `peerMetadata`, it's no longer in `ConnectionOptions`

Question: should we finally create a context and a provider for this lib and manage the socket etc there, in one place? It will be problematic if someone calls `useMembraneServer` twice.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -175,6 +175,6 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:${versions.coroutines}"
-  implementation 'com.github.jellyfish-dev:membrane-webrtc-android:4.5.0'
+  implementation 'com.github.jellyfish-dev:membrane-webrtc-android:c9a84a5280365a2874911e936349d7cfbdbef1b1'
   api 'com.github.davidliu:audioswitch:8edf84ee46cdbc84c2b7725aa8f8d66363e19876'
 }
--- a/android/src/main/java/com/reactnativemembrane/MembraneModule.kt
+++ b/android/src/main/java/com/reactnativemembrane/MembraneModule.kt
@@ -1,21 +1,20 @@
 package com.reactnativemembrane
 
 import android.app.Activity
-import android.content.Context
 import android.content.Intent
-import android.media.AudioManager
 import android.media.projection.MediaProjectionManager
 import androidx.appcompat.app.AppCompatActivity
 import com.facebook.react.bridge.*
 import com.facebook.react.modules.core.DeviceEventManagerModule
 import com.twilio.audioswitch.AudioDevice
-import kotlinx.coroutines.Dispatchers
-import org.membraneframework.rtc.*
+import org.membraneframework.rtc.MembraneRTC
+import org.membraneframework.rtc.MembraneRTCListener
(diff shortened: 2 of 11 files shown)


android/src/main/java/com/reactnativemembrane/MembraneModule.kt
```diff
@@ -130,76 +122,125 @@ class MembraneModule(reactContext: ReactApplicationContext) :
...
-
-    if(connectionOptions.hasKey("videoTrackEnabled"))
-      this.isCameraOn = connectionOptions.getBoolean("videoTrackEnabled")
-    if(connectionOptions.hasKey("audioTrackEnabled"))
-      this.isMicrophoneOn = connectionOptions.getBoolean("audioTrackEnabled")
-
-    val socketConnectionParams = connectionOptions.getMap("connectionParams")?.toMap() ?: mutableMapOf()
+  fun create(url: String, createOptions: ReadableMap, promise: Promise) {
+    val videoQuality = createOptions.getString("quality")
+    var flipVideo = true
```
skyman503, line 127: What about `flipVideo = if (createOptions.hasKey("flipVideo")) createOptions.getBoolean("flipVideo") else true` ?

android/build.gradle
```diff
@@ -175,6 +175,6 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:${versions.coroutines}"
-  implementation 'com.github.jellyfish-dev:membrane-webrtc-android:4.5.0'
+  implementation 'com.github.jellyfish-dev:membrane-webrtc-android:c9a84a5280365a2874911e936349d7cfbdbef1b1'
```
skyman503, line 178: Mby we should wait for the android and ios release before merging this.

android/src/main/java/com/reactnativemembrane/MembraneModule.kt
```diff
@@ -130,76 +122,125 @@ class MembraneModule(reactContext: ReactApplicationContext) :
...
+      "QHD169" -> VideoParameters.presetQHD169
+      "HD169" -> VideoParameters.presetHD169
+      "FHD169" -> VideoParameters.presetFHD169
+      "QVGA43" -> VideoParameters.presetQVGA43
+      "VGA43" -> VideoParameters.presetVGA43
+      "QHD43" -> VideoParameters.presetQHD43
+      "HD43" -> VideoParameters.presetHD43
+      "FHD43" -> VideoParameters.presetFHD43
+      else -> VideoParameters.presetVGA169
+    }
```
skyman503, line 164: What about extracting it to a new function? This one is already pretty big

src/index.tsx
```diff
@@ -349,12 +346,24 @@ export function useMembraneServer() {
   const [error, setError] = useState<string | null>(null);
   // prevent user from calling connect methods multiple times
   const lock = useRef(false);
+  // phoenix sockets are not ts typed :(
```
skyman503, line 349: huh, that's annoying.

skyman503 (changes requested):

skyman503 (approved):
Looks good! thanks

PR #107, opened 2023-05-16
Fix camera on/off switch in Room screen

graszka22:
This is pretty tricky bug :D It was caused by this commit: https://github.com/jellyfish-dev/react-native-membrane-webrtc/pull/93/files
Here is what happens when we go to the room screen and toggle the camera:
- in preview screen everything is ok, you can turn the camera on/off, no problem
- you go to the room screen
- the preview screen is in the stack but it's detached by the navigation
- the preview view is detached too, the local track is disposed all ok
- the room screen is attached and the local track is initialized again
- then we turn the camera off, the local track is disabled
- hey, remember that preview screen? it's not attached but it's still somewhere in the memory and it just updated and replaced the preview view with a placeholder - it's nothing bad (yet!)
- then we turn the camera back on
- the preview screen is updated again and wants to create a preview view, the `initialize` function is called even though it's not attached to window and we can't have two local camera tracks so weird stuff happens

Solution? Now we initialize the track only when window is attached and it seems to be fixed now.

### DIFF
--- a/android/src/main/java/com/reactnativemembrane/VideoPreviewView.kt
+++ b/android/src/main/java/com/reactnativemembrane/VideoPreviewView.kt
@@ -8,40 +8,41 @@ import org.membraneframework.rtc.media.VideoParameters
 import org.webrtc.EglBase
 
 class VideoPreviewView(private val context: Context): VideoTextureViewRenderer(context) {
-  private lateinit var localVideoTrack: LocalVideoTrack
-  private lateinit var eglBase: EglBase
-  private lateinit var peerConnectionFactory: PeerConnectionFactory
+  private var localVideoTrack: LocalVideoTrack? = null
+  private var eglBase: EglBase? = null
+  private var peerConnectionFactory: PeerConnectionFactory? = null
   private var isInitialized: Boolean = false
 
-  init {
-    initialize()
-  }
+  private var captureDeviceId: String? = null
 
   private fun initialize() {
     if(isInitialized) return
     isInitialized = true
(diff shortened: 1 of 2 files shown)


chanphiromsok:
I wonder why this package does not stay in membrane organize?

skyman503 (approved):
Works fine now, thanks a lot for that

mickel8:
@chanphiromsok we haved moved everything related to Jellyfish to a new organization. This includes RTC Engine and its SDKs. Mostly because they differ a lot from a typical Membrane plugin, they are some kind of prodcut built on top of Membrane and it's easier to document, communicate, etc around a separate organization

chanphiromsok:
> @chanphiromsok we haved moved everything related to Jellyfish to a new organization. This includes RTC Engine and its SDKs. Mostly because they differ a lot from a typical Membrane plugin, they are some kind of prodcut built on top of Membrane and it's easier to document, communicate, etc around a separate organization

so does this package support my ow membrane webrtc server ?

graszka22:
@chanphiromsok yes, if you're using RTC Engine in your server, this package should work. You don't need to use Jellyfish at all. 
In particular, in our example Videroom app we're not using any Jellyfish stuff at all (yet).

chanphiromsok:
> @chanphiromsok yes, if you're using RTC Engine in your server, this package should work. You don't need to use Jellyfish at all. In particular, in our example Videroom app we're not using any Jellyfish stuff at all (yet).

thanks you so much

PR #116, opened 2023-05-30
[MV-468] Fix standard button working multiple times when clicked very fast

skyman503:


example/src/shared/debouncer.tsx
```diff
@@ -1,12 +1,16 @@
 import { debounce } from 'lodash';
 import { useCallback } from 'react';
 
+// Make sure that the function used as a callback stays
+// always the same since makaing changes to it won't
```
graszka22, line 5:
```suggestion
// always the same since making changes to it won't
```

graszka22 (approved):

PR #117, opened 2023-06-01
[MV-469] Add voice activity indicator UI

graszka22:


skyman503 (commented):
Looks good to me however according to the designs, if pinned user is talking they should also be given a green border which is not happening.

graszka22:
> Looks good to me however according to the designs, if pinned user is talking they should also be given a green border which is not happening.

Good catch! Fixed 👍

skyman503 (approved):

PR #119, opened 2023-06-06
[MV-471] Fix joining with muted microphone

graszka22:
The debounce bug comes again :D 
This time I hope I fixed it for good. Joining with muted micro should work now and the issue fix opening double preview appears to be still fixed (but please check that!)


skyman503:
Seems to be working fine now, thanks. The only problem I could find is when you click fast on the initial screen simultaneously on the create and join button then 2 screens will be mounded, but I think it's not in scope of this issue

skyman503 (approved):

PR #120, opened 2023-06-12
[MV-470] Always show currently talking users

skyman503:


example/src/shared/utils.ts
```diff
@@ -25,3 +25,26 @@ export const checkIfUrl = (val: string) => {
 export const extractRoomNameFromUrl = (url: string) => {
   return url.substring(url.indexOf(VIDEOROOM_URL) + VIDEOROOM_URL.length);
 };
+
+export const checkIfArraysAreTheSame = (
```
graszka22, line 29: https://lodash.com/docs/4.17.15#isEqual ?

example/src/screens/Room.tsx
```diff
@@ -32,20 +44,130 @@ export const Room = ({ navigation }: Props) => {
   const participants = Membrane.useRoomParticipants();
   const [focusedParticipantData, setFocusedParticipantData] =
     useState<Participant | null>(null);
+  // This string array will containg strings in a form of participantId+trackId
+  // to differentiate between camera and screenshare video tracks.
+  const [participantsOrder, setParticipantsOrder] = useState<string[] | null>(
+    null
+  );
+
+  const getFirstNotSpeakingVisablePlace = (
```
graszka22, line 53:
```suggestion
  const getFirstNotSpeakingVisiblePlace = (
```

example/src/screens/Room.tsx
```diff
@@ -32,20 +44,130 @@ export const Room = ({ navigation }: Props) => {
...
+  };
+
+  const saveCurrentOrder = (participants: ParticipantWithTrack[]) => {
+    const newOrder: string[] = [];
+
+    for (let index = 0; index < participants.length; index++) {
+      newOrder.push(
+        participants[index].participant.id + participants[index].trackId
+      );
+    }
```
graszka22, line 83:
```suggestion
const newOrder = participants.map(p => p.participant.id+p.trackId)
```

example/src/screens/Room.tsx
```diff
@@ -32,20 +44,130 @@ export const Room = ({ navigation }: Props) => {
   const participants = Membrane.useRoomParticipants();
   const [focusedParticipantData, setFocusedParticipantData] =
     useState<Participant | null>(null);
+  // This string array will containg strings in a form of participantId+trackId
+  // to differentiate between camera and screenshare video tracks.
+  const [participantsOrder, setParticipantsOrder] = useState<string[] | null>(
```
graszka22, line 49: are you sure this should be in state and not in ref? setting state when rendering sounds bad to me

graszka22 (changes requested):

example/src/shared/utils.ts
skyman503: I was looking at docs and couldn't find this function..., will use it now. thanks

example/src/shared/utils.ts
```diff
@@ -25,3 +25,26 @@ export const checkIfUrl = (val: string) => {
...
+
+  for (let index = 0; index < arr1.length; index++) {
+    if (arr1[index] !== arr2[index]) {
+      return false;
+    }
+  }
+  return true;
+};
+
+export const getNumberOfCurrentlyVisiblePlaces = (
```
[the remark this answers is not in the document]
skyman503: Should be fixed now

example/src/screens/Room.tsx
graszka22: when vad status changes, the room will re-render anyway, because `participants` array (the one we got from `useRoomParticipants` hook) changes, right? so I think we're making unecessary re-render here:
1) order is kept in the state
- participants are initially rendered in the old order
- in render the new order is set
- room is rendered again with the new order
2) order is kept in the ref
- the new order (participants array) is calculated based on current vad status and data in `participantsOrder` and the participants are rendered in the new order on the first time, no additional re-render is needed

at least that's how I understand your code, it's complicated, let me know what do you think 😅

skyman503: Yeah, prob. it can be done with useRef, if we want to spare that one render

example/src/screens/Room.tsx
```diff
@@ -32,20 +41,47 @@ export const Room = ({ navigation }: Props) => {
   const participants = Membrane.useRoomParticipants();
   const [focusedParticipantData, setFocusedParticipantData] =
     useState<Participant | null>(null);
+  const { applyPrevOrder, orderAndSave, lru } = ParticipantsTracksWindowManager(
```
graszka22, line 44: it looks like `ParticipantsTracksWindowManager` is a hook, in that case it should use `useSomething` convention
also "window manager"? o.o what windows are you managing?

example/src/screens/Room.tsx
```diff
@@ -32,20 +41,47 @@ export const Room = ({ navigation }: Props) => {
...
-          .map((t) => ({
-            participant: p,
-            trackId: t.id,
-          }));
-      }
-      return { participant: p };
-    })
-    .flat();
+  // LRU initialization
+  useEffect(() => {
```
graszka22, line 49: shouldn't you put this in the `ParticipantsTracksWindowManager`? it's a hook, it can have `useEffect`

example/src/shared/participantsWindowManager.ts
```diff
@@ -0,0 +1,212 @@
...
+  isActive: boolean;
+  index: number;
+  // Local user and screenshare tracks
+  // are not movable and should stay on top.
+  isMovable: boolean;
+};
+
+// Used to manage order in which participants tracks are displayed.
+export const ParticipantsTracksWindowManager = (
+  maxNumUsersOnScreen: number
```
graszka22, line 21: you can import that, it's not necessary to use a parameter?

example/src/shared/participantsWindowManager.ts
```diff
@@ -0,0 +1,212 @@
...
+    b: ParticipantWithTrack
+  ) => {
+    if (a.participant.type === 'Local') {
+      return -1;
+    } else if (b.participant.type === 'Local') {
+      return 1;
+    }
+
+    if (
+      a.participant.tracks.find((t) => t.id === a.trackId)?.metadata.type ===
```
graszka22, line 54: you can add a function that for given participant checks if it's a screenshare?

example/src/screens/Room.tsx
```diff
@@ -32,20 +41,47 @@ export const Room = ({ navigation }: Props) => {
...
+        isMovable: true,
+      });
+    }
+
+    return () => {
+      lru.current = [];
+    };
+  }, []);
+
+  const orderParticipantsAccordingToVadStatus = (
```
graszka22, line 64: this function can be put in `ParticipantsTracksWindowManager` and returned, instead of `orderAndSave` and `applyPrevOrder`?

graszka22 (changes requested):
please try to debug the black squares bug
and your ordering hook is a really good candidate to test ^^ because it doesn't really render anything, just pure algorithms
you can also consider a hook that instead of returning a function that orders participant, returns already ordered and processed participants with tracks etc.
you can add tests in another PR

skyman503:
> please try to debug the black squares bug and your ordering hook is a really good candidate to test ^^ because it doesn't really render anything, just pure algorithms you can also consider a hook that instead of returning a function that orders participant, returns already ordered and processed participants with tracks etc. you can add tests in another PR

Yep, I was planning to add tests in the followup. As for the black squares at this point I'm not entirely sure what causes that, will investigate further.

example/src/screens/Room.tsx

example/src/shared/participantsWindowManager.ts
skyman503: Done.

example/src/screens/Room.tsx

PR #122, opened 2023-06-15
[MV-480] remove speaking information from screenshare tracks

skyman503:
Please check scenario: Remote participants stars screensharing and then talking. No vad status information should be displayed on the screenshare video track (both when focused and not)

### DIFF
--- a/example/src/components/FocusedParticipant.tsx
+++ b/example/src/components/FocusedParticipant.tsx
@@ -26,14 +26,17 @@ export const FocusedParticipant = ({
 
   const isParticipantSpeaking =
     focusedParticipant.participant.tracks.find((t) => t.type === 'Audio')
-      ?.vadStatus === 'speech' &&
-    focusedTrack?.metadata.type !== 'screensharing';
+      ?.vadStatus === 'speech';
+
+  const isScreenshare = focusedTrack?.metadata.type !== 'screensharing';
 
   return (
     <View
       style={[
         styles.focusedParticipantContainer,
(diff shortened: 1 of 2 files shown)


example/src/components/Participants.tsx
```diff
@@ -75,7 +82,7 @@ export const Participants = ({ participants, onPress }: ParticipantsProp) => {
               key={p.participant.id + p.trackId}
               style={[
                 getStylesForParticipants(),
-                isParticipantSpeaking(p)
+                isParticipantSpeaking(p) && isVideoTrackScreenshare(p) === false
```
graszka22, line 85: it would be better to include check for screensharing to `isParticipantSpeaking`

example/src/components/FocusedParticipant.tsx
```diff
@@ -28,11 +28,15 @@ export const FocusedParticipant = ({
     focusedParticipant.participant.tracks.find((t) => t.type === 'Audio')
       ?.vadStatus === 'speech';
 
+  const isScreenshare = focusedTrack?.metadata.type === 'screensharing';
```
graszka22, line 31: add it to `isParticipantSpeaking` too

example/src/components/Participants.tsx
skyman503: Done.

example/src/components/FocusedParticipant.tsx
skyman503: Done.

graszka22 (approved):

PR #123, opened 2023-06-19
[MV-470] Add tests for the participants ordering algorithm

skyman503:
Additionally fixes a bug with screencast tracks order that was detected by new tests


graszka22 (approved):

graszka22:
The last test could test more, for example if its ok if a participant leaves and then someone speaks, then the test could check if the rlu behaves right.
But overall ok 👍

PR #125, opened 2023-06-21
[MV-482] Fix function types in mobile app

skyman503:


src/index.tsx
```diff
@@ -333,6 +333,8 @@ export type RTCInboundStats = {
 
 export type RTCStats = RTCOutboundStats | RTCInboundStats;
 
+type RTCStatsObject = { [key: string]: RTCStats };
```
graszka22, line 336: I don't like this naming. Can you tell the difference between `RTCStats` and `RTCStatsObject` just by looking at the name? `RTCStats` is also an object, right? 
Maybe change `RTCStats` to something like `RTCTrackStats` and `RTCStatsObject` to `RTCStats`? That way we know that the first one is one track's stats and the second are all stats. But renaming `RTCStats` also isn't cool, so if you have any better ideas go ahead

skyman503: Yeah, I'm not sure what to do with naming either.

graszka22 (approved):

PR #126, opened 2023-06-22
Fix memory leak

graszka22:
To fix completely: https://github.com/jellyfish-dev/membrane-webrtc-ios/pull/43 
To test:
- make a lot of remote tracks (or display one track many times)
- turn them on/off several times
- the app memory in xcode profiler goes up, it should stay on approximately the same level


skyman503 (approved):
🆒

PR #127, opened 2023-06-22
Bump deps

graszka22:


ios/Membrane.swift
```diff
@@ -185,6 +185,8 @@ class Membrane: RCTEventEmitter, MembraneRTCDelegate {
     let localEndpointId = UUID().uuidString
     self.localEndpointId = localEndpointId
     
+    self.localUserMetadata = (connectionOptions["endpointMetadata"] as? NSDictionary)?.toMetadata() ?? Metadata()
```
skyman503, line 188: Shouldn't the similar change be done to the MembraneModule.kt ?

graszka22: it's ok, in `MembraneModule.kt` the local metadata is set in the connect function and it doesn't really matter if we do this in create or connect
and I tested on Android and it works (it was crashing on ios if we turned off the camera)

skyman503 (approved):
😃

PR #129, opened 2023-07-10
[EX-49] Move `useSimulcast` hook logic to `useCamera` hook in react native webrtc sdk

skyman503:


src/index.tsx
```diff
@@ -532,6 +545,49 @@ export function useCamera() {
...
+  /**
+   * sets track encoding that server should send to the client library.
+   * The encoding will be sent whenever it is available. If choosen encoding is
+   * temporarily unavailable, some other encoding will be sent until choosen encoding
+   *  becomes active again.
+   *
+   * @param trackId id of a track which encoding you want to select
+   * @param encoding encoding to select
+   */
+  const setTargetTrackEncoding = useCallback(
```
graszka22, line 557: this one is not related to the local video track, so it would be better to move it to separate function?

graszka22 (changes requested):

src/index.tsx
skyman503: I placed it in a new hook.

src/index.tsx
```diff
@@ -518,11 +518,46 @@ export function useEndpoints() {
...
```
graszka22, line 534: no need for a hook, just function is enough 
see for example `updateEndpointMetadata` function

graszka22 (changes requested):

graszka22 (changes requested):
also `setVideoTrackBandwidth` is related to the camera track, we can also put it in the `useCamera` hook

src/__mocks__/index.tsx
```diff
@@ -77,6 +81,7 @@ export const useScreencast = () => {
     simulcastConfig: NOOP,
     setScreencastTrackEncodingBandwidth: NOOP,
     setScreencastTrackBandwidth: NOOP,
+    setVideoTrackBandwidth: NOOP,
```
graszka22, line 84: wait, it's not in useScreeencast 🤔

graszka22 (changes requested):

graszka22 (approved):

src/__mocks__/index.tsx
skyman503: yeah, good catch, I always mix up simulcast and screencast

PR #130, opened 2023-07-10
[EX-29] Screencast crash in mobile app on Android

skyman503:


skyman503:
@graszka22  PTAL at this proposal of a solution for the race condition when dealing with triggering screenshare

android/src/main/java/com/reactnativemembrane/MembraneModule.kt
```diff
@@ -384,6 +388,9 @@ class MembraneModule(reactContext: ReactApplicationContext) :
         getSimulcastConfigFromOptions(screencastOptions)
       this@MembraneModule.screencastMaxBandwidth = getMaxBandwidthFromOptions(screencastOptions)
       screencastPromise = promise
+
+      screencastMutex.lock()
+
       if (!isScreenCastOn) {
         if (!ensureConnected(promise)) return@launch
```
graszka22, line 395: I doubt we ever get here but we should unlock the mutex here just in case

PR #131, opened 2023-07-10
Expo modules

graszka22:
With this our sdk supports fabric architecture. How to test:
- For Android: in the `gradle.properties` file there is a variable `newArchEnabled` - set it to true to enable fabric architecture for example app (and therefore our sdk) HOWEVER there is a problem on some newer phones that on the second screen you can't click anything and it's probably a bug in react-navigation lib, it's being fixed by the open-source team
 
- For iOS: in the `example/ios/Podfile.properties.json` there is a variable `newArchEnabled`, set it to true to enable fabric architecture for example app (and therefore our sdk) HOWEVER there is a bug in expo-font library and similar and all fonts and icons are broaken on fabric, expo team is working on this

Also remember when switching between architectures to always clean cache etc. you may also need to delete `Pods` or `build` folders or even just clone the repo again :p 

When reviewing focus on the new modules (whether the code style is ok) and whether I've missed something (this repo is recreated from scratch 😅).
Also react-native and other libraries are updated <3

Thoughts:
- do we want a ci flow to check if example for fabric and paper compiles?


graszka22:
@skyman503 If you have a while you can take a look at this PR, it shouldn't change much. Things that are not working yet:
- app on fabric crashes on ios (must be related to rn 71 -> 72 update or related updates)
- tests are broken, because we have missing dev libraries (react, react-native and react-dom) in package.json, however when I install them the app crashes because we have double react package 😅 

Focus on the code in modules, those thousands lines of changes are mostly yarn.locks, generated boilerplate etc.

android/src/main/java/com/reactnativemembrane/MembraneWebRTCModule.kt
```diff
@@ -0,0 +1,279 @@
...
+
+    @Field
+    val maxBandwidthInt: Int = 0
+}
+
+class MembraneWebRTCModule : Module() {
+    override fun definition() = ModuleDefinition {
+        Name("MembraneWebRTC")
+
+        Events(
```
skyman503, line 78: What about moving that bug fix from ex-29 here, since after this changes are applied I'd have to redo my PR. So wdyt about mimicing my changes here, since it is pretty short and straight forward ?

example/src/model/VideoroomContext.tsx
```diff
@@ -160,6 +160,8 @@ const VideoroomContextProvider = ({ children }: VideoroomContextProps) => {
 
   useEffect(() => {
     if (error) {
+      console.log('ERR', error);
```
skyman503, line 163: needed?

example/src/screens/Preview.tsx
```diff
@@ -65,6 +65,8 @@ export const Preview = ({ navigation, route }: Props) => {
       await connectAndJoinRoom();
       navigation.navigate('Room');
     } catch (err) {
+      console.log('ERR2', err);
```
skyman503, line 68: same here

src/utils.ts
```diff
@@ -0,0 +1,4 @@
+export function isJest() {
+  // @ts-ignore
```
skyman503, line 2: still needed?

skyman503 (changes requested):

skyman503:
As for the question in description in my opinion, as long as we plan on supporting both platforms ci should also take care of both of them.

example/ios/reactnativemembranewebrtcexample.xcodeproj/xcshareddata/xcschemes/reactnativemembranewebrtcexample.xcscheme
```diff
@@ -29,9 +15,9 @@
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "13B07F861A680F5B00A75B9A"
-               BuildableName = "MembraneExample.app"
-               BlueprintName = "MembraneExample"
-               ReferencedContainer = "container:MembraneExample.xcodeproj">
+               BuildableName = "reactnativemembranewebrtcexample.app"
```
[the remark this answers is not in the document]
graszka22: The reason is that in `MembraneExample.xcodeproj` and `reactnativemembranewebrtcexample.xcodeproj` there are files that are gitignored, when switching between branches those gitignored fils are left by git

so my suggestion would be to just remove those folders when switching branches or rename reactnativemembranewebrtcexample back to MembraneExample but I don't know if it's a good idea to touch it 😅

src/utils.ts
graszka22: yes

graszka22:
> As for the question in description in my opinion, as long as we plan on supporting both platforms ci should also take care of both of them.

ok, I will create a task to add ci for them

skyman503 (approved):
:yolo:

PR #132, opened 2023-07-11
[EX-29] Screencast crash in mobile app

skyman503:
After this changes are applied screencast is always stopped before leaving the call, which fixes the bug.

### DIFF
--- a/ios/Membrane.swift
+++ b/ios/Membrane.swift
@@ -248,6 +248,15 @@ class Membrane: RCTEventEmitter, MembraneRTCDelegate {
   
   @objc(disconnect:withRejecter:)
   func disconnect(resolve:RCTPromiseResolveBlock, reject:RCTPromiseRejectBlock) -> Void {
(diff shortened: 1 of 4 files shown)


ios/Membrane.swift
```diff
@@ -248,6 +248,15 @@ class Membrane: RCTEventEmitter, MembraneRTCDelegate {
   
   @objc(disconnect:withRejecter:)
   func disconnect(resolve:RCTPromiseResolveBlock, reject:RCTPromiseRejectBlock) -> Void {
+    guard isScreensharingEnabled == false else {
+      DispatchQueue.main.async {
+        RPSystemBroadcastPickerView.show(for: screencastExtensionBundleId)
```
graszka22, line 253: this only shows the dialog to the user, an evil user still might choose to keep the screencast on and it's a bad ux
I know the screencast is controlled by the broadcast extension and probably it's the only way to stop the broadcast but have you tried maybe calling `stop` on the screencast track?

PR #134, opened 2023-07-28
[EX-108] Add generic types instead of any

skyman503:


example/src/types/MetadataTypes.ts
```diff
@@ -0,0 +1,8 @@
+export type TrackMetadataType = {
```
graszka22, line 1: do we need to add `Type` suffix here? isn't it obvious that a type is a type?

src/index.tsx
```diff
@@ -196,18 +196,18 @@ export function useWebRTC() {
  * This hook provides live updates of room endpoints.
  * @returns An array of room endpoints.
  */
-export function useEndpoints() {
-  const [endpoints, setEndpoints] = useState<Endpoint[]>([]);
+export function useEndpoints<T extends Metadata, S extends Metadata>() {
```
graszka22, line 199: can't T and S be more descriptive? I don't know what do they mean

graszka22: also probably the metadata can be different for different tracks

for example, an audio track can have 'muted' property, which doesn't make sense for camera track

src/index.tsx
```diff
@@ -508,22 +510,26 @@ export function useScreencast() {
...
+export async function updateEndpointMetadata<T extends Metadata>(metadata: T) {
   await MembraneWebRTCModule.updateEndpointMetadata(metadata);
 }
 
 /**
  * a function that updates video metadata on the server.
  * @param metadata a map string -> any containing video track metadata to be sent to the server
  */
-export async function updateVideoTrackMetadata(metadata: Metadata) {
+export async function updateVideoTrackMetadata<T extends Metadata>(
```
graszka22, line 521: there is a problem with that: I can specify different type for metadata when starting the camera and when updating camera track metadata
ideally it'd be inferred automatically but I don't know how to do this or if it's possible (maybe with some api changes?)

graszka22 (commented):
I added some comments, treat them rather like proposals, some of them maybe doesn't have sense ;) 
in general let's also think if adding those types improved user experience or it's just an additional hassle

example/src/types/MetadataTypes.ts
skyman503: Removed.

graszka22 (approved):
LGTM 👍 
Also let Kamil take a look

skyman503:
@kamil-stasiak  PTAL

kamil-stasiak (approved):

PR #135, opened 2023-07-28
[EX-18] Add sound detection

incubo4u:


example/package.json
```diff
@@ -33,6 +33,7 @@
     "react-native-reanimated": "~3.3.0",
     "react-native-safe-area-context": "4.6.3",
     "react-native-screens": "~3.22.0",
+    "react-native-sound-level": "^1.2.0",
```
skyman503, line 36: Remove

src/index.tsx
```diff
@@ -383,6 +389,20 @@ export function useMicrophone() {
     setIsMicrophoneOn(MembraneWebRTCModule.isMicrophoneOn);
     return () => eventListener.remove();
   }, []);
+  /**
+   * This sets the voice activity detection state.
+   */
+  useEffect(() => {
+    if (isVADOn) {
+      RNSoundLevel.onNewFrame = (data: VADData) => {
```
graszka22, line 397: why are we doing it in useEffect, if we do it only when vad is started and there is no cleanup?
I see two options to do it better:
1. Make only `startVADMonitor` and `stopVADMonitor`. `startVADMonitor` sets the `onNewFrame` callback and calls `RNSoundLevel.start`. User is responsible for calling `stopVADMonitor` when they're done with it, but they have more control over it and it's simpler. Also you can get rid of `isVADOn` state.
2. Create a separate hook with `useEffect` (for example `useVADMonitor`). When used, the hook starts up RNSoundLevel and sets up callback (both in useEffect). When no longer used, it cleans up automatically (in `useEffect` return function). It's easier to use, but user has less control (what happens when they use the hook in two different places in the app?).

src/index.tsx
```diff
@@ -383,6 +389,20 @@ export function useMicrophone() {
...
+        const dB = data.value;
+        const speechThreshold = -60;
```
graszka22, line 399: I'd make that an option, maybe users want different sensitivity or something

src/index.tsx
```diff
@@ -403,8 +423,35 @@ export function useMicrophone() {
...
+    RNSoundLevel.stop();
+    setIsVADOn(false);
+  }, []);
 
-  return { isMicrophoneOn, toggleMicrophone, startMicrophone };
+  return {
+    isMicrophoneOn,
+    isSpeaking,
+    toggleMicrophone,
+    startVADMonitor,
```
graszka22, line 451: please add it also in the example app, maybe in dev mode (long press on membrane logo)

graszka22 (changes requested):
One more thing: you're going to ~copy~ heavily inspire from `react-native-sound-level` to bring this functionality into the native sdks (android and ios). Can RN side use the functionality from our native sdks instead of adding a dependency on `react-native-sound-level`? It's the same code, but one less dependency and less duplication and easier maintenance :3 

Also please set the PR title to something more descriptive

PR #145, opened 2023-11-14
Incubo4u/ex-228/mapping Bug fix

incubo4u:


example/src/components/RoomParticipant.tsx
```diff
@@ -56,12 +56,8 @@ export const RoomParticipant = ({
...
   const buttonOpacity = useSharedValue(0);
 
-  const participantHasVideo = () => {
-    if (videoTrack) {
-      return videoTrack.metadata.active;
-    }
-    return false;
-  };
+  const participantHasVideo = () =>
+    videoTrack ? videoTrack.metadata.active : false;
```
karkakol, line 60: videoTrack && videoTrack.metadata.active does the same but shorter :)

ios/MembraneWebRTCModule.swift
```diff
@@ -104,7 +104,7 @@ public class MembraneWebRTCModule: Module {
       try membraneWebRTC.toggleMicrophone()
     }
     
-    Property("IsCameraOn") {
+    Property("isCameraOn") {
```
karkakol, line 107: Why here camelCase and in android PascalCase?

src/index.tsx
```diff
@@ -398,12 +398,14 @@ export function useCamera() {
...
  */
 export function useMicrophone() {
-  const [isMicrophoneOn, setIsMicrophoneOn] = useState<boolean>(false);
+  const [isMicrophoneOn, setIsMicrophoneOn] = useState<boolean>(
+    MembraneWebRTCModule.isMicrophoneOn
+  );
 
   useEffect(() => {
     const eventListener = eventEmitter.addListener<IsMicrophoneOnEvent>(
       'IsMicrophoneOn',
```
karkakol, line 407: I was wondering is it a good idea to create a script which generates swift, kotlin and react  files based on json config, that will get rid off discrepancy between events naming

config could look like:
{
  "MAMA" : "MAMA",
  "TATA" : "TATA",
  "BRAT" : "BRAT"
}
 and script ( in python for example):
 
 import json

with open('variables.json', 'r') as f:
    variables = json.load(f)

with open('variables.ts', 'w') as f:
    for key, value in variables.items():
        f.write(f'const {key} = "{value}";\n')

with open('variables.swift', 'w') as f:
    for key, value in variables.items():
        f.write(f'let {key} = "{value}"\n')

with open('variables.kt', 'w') as f:
    f.write('object Constants {\n')
    for key, value in variables.items():
        f.write(f'    val {key} = "{value}"\n')
    f.write('}\n')

android/src/main/java/com/reactnativemembrane/MembraneWebRTCModule.kt
```diff
@@ -193,19 +193,19 @@ class MembraneWebRTCModule : Module() {
...
         }
 
         AsyncFunction("updateAudioTrackMetadata") Coroutine { metadata: Map<String, Any> ->
             withContext(Dispatchers.Main) {
-                membraneWebRTC.updateAudioTrackMetadata(metadata)
+                membraneWebRTC.updateLocalAudioTrackMetadata(metadata)
             }
         }
 
         AsyncFunction("updateScreencastTrackMetadata") Coroutine { metadata: Map<String, Any> ->
```
karkakol, line 206: Are you sure that discrepancy between AsyncFunction name and mambraneWebRTC.method is valid?

karkakol (changes requested):

ios/MembraneWebRTCModule.swift
incubo4u: https://github.com/jellyfish-dev/react-native-membrane-webrtc/blob/2184aa99fd8bf36207b9e4d31e66a4b3fc56badb/android/src/main/java/com/reactnativemembrane/MembraneWebRTCModule.kt#L144

i belive that on android camelCase is used

android/src/main/java/com/reactnativemembrane/MembraneWebRTCModule.kt
incubo4u: Are you asking if the functions can have different names or if they should have different names? They can indeed have different names, and the function will still be correctly called. However, I agree that they should ideally have the same name. I opted to change the name for clarity in the native code and didn't want to introduce a breaking change to React Native

incubo4u: If you believe this is a bad idea, please let me know 😊

ios/MembraneWebRTC.swift
```diff
@@ -8,840 +8,883 @@ import ExpoModulesCore
...
+                    return
+                }
+                do {
+                    guard let screencastTrack = localScreencastTrack else {return }
+                    try removeTrackFromLocalEndpoint(screencastTrack)
+                    localScreencastTrack = nil
+                    try setScreencastTrackState(isEnabled: false)
+                }
+                catch{
+                    print("Error stoping screencast: \(error)")
```
karkakol, line 442: why print and not logger?

ios/MembraneWebRTC.swift
```diff
@@ -8,840 +8,883 @@ import ExpoModulesCore
...
+        try ensureConnected()
+        guard let screencastExtensionBundleId = Bundle.main.infoDictionary?["ScreencastExtensionBundleId"] as? String else {
+            throw Exception(name: "E_NO_BUNDLE_ID_SET", description: "No screencast extension bundle id set. Please set ScreencastExtensionBundleId in Info.plist")
+        }
+        guard let appGroupName = Bundle.main.infoDictionary?["AppGroupName"] as? String else {
+            throw Exception(name: "E_NO_APP_GROUP_SET", description: "No app group name set. Please set AppGroupName in Info.plist")
+        }
+        // if screensharing is enabled it must be closed by the Broadcast Extension, not by our application
+        // the only thing we can do is to display stop recording button, which we already do
+        guard isScreensharingEnabled == false else {
```
karkakol, line 398: why not !isScreensharingEnabled ?

ios/MembraneWebRTC.swift
```diff
@@ -1,844 +1,883 @@
...
+    }
+    private func removeTrackFromLocalEndpoint(_ track: LocalVideoTrack) throws {
+        try ensureEndpoints()
+        if let localEndpointId = localEndpointId {
+            guard let localEndpoint = MembraneRoom.sharedInstance.endpoints[localEndpointId] else { return }
+            MembraneRoom.sharedInstance.endpoints[localEndpointId] = localEndpoint.removeTrack(trackId: track.trackId())
+            emitEndpoints()
+        }
+    }
+    private func removeTrackFromLocalEndpoint(_ track: LocalAudioTrack){}
```
kamil-stasiak, line 331: Why is this function empty?

android/src/main/java/com/reactnativemembrane/MembraneWebRTC.kt
```diff
@@ -609,23 +655,23 @@ class MembraneWebRTC(val sendEvent: (name: String, data: Map<String, Any?>) -> U
...
-        audioDevices: List<AudioDevice>,
-        selectedDevice: AudioDevice?
+        audioDevices: List<AudioDevice>, selectedDevice: AudioDevice?
     ) {
-        val map = mapOf(
-            "selectedDevice" to (if (selectedDevice != null) audioDeviceAsRNMap(selectedDevice) else null),
-            "availableDevices" to audioDevices.map { audioDevice -> audioDeviceAsRNMap(audioDevice) }
-        )
+        val map =
+            mapOf("selectedDevice" to (if (selectedDevice != null) audioDeviceAsRNMap(selectedDevice) else null),
```
kamil-stasiak, line 661: The main key of this map is `selectedDevice` and the event id is `AudioDeviceUpdate`. Shouldn't they be the same?

kamil-stasiak (changes requested):
This code is readeble

ios/MembraneWebRTC.swift
```diff
@@ -8,837 +8,883 @@ import ExpoModulesCore
...
+    //  func connect(metadata: [String: Any], promise: Promise) -> Void {
+    //      connectPromise = promise
+    //      .localUserMetadata = metadata.toMetadata()
+    //
+    //      if let localEndpointId = localEndpointId {
+    //          .endpoints[localEndpointId]?.metadata = metadata.toMetadata()
+    //    }
+    //
+    //    membraneRTC?.connect(metadata: metadata.toMetadata())
+    //  }
```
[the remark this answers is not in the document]
incubo4u: nope, removing

android/src/main/java/com/reactnativemembrane/MembraneWebRTC.kt
incubo4u: there are 2 main keys in this map

kamil-stasiak (approved):

example/src/components/RoomParticipant.tsx
```diff
@@ -139,11 +135,11 @@ export const RoomParticipant = ({
     }
     return null;
   };
-
   return (
     <View style={styles.fill}>
       <Pressable onPress={triggerShowingPinButton} style={styles.fill}>
         {participantHasVideo() ? (
+          // @ts-ignore
```
karkakol, line 142: czemu ts-ignore?
barfdo nieładnie

example/src/screens/Preview.tsx
```diff
@@ -103,6 +103,7 @@ export const Preview = ({ navigation, route }: Props) => {
 
             <View style={styles.cameraPreview}>
               {isCameraOn ? (
+                // @ts-ignore
```
karkakol, line 106: dlaczemu ts-ignore?

PR #162, opened 2024-04-23
stop screencast when app killed on android

karkakol:


android/src/main/java/com/reactnativemembrane/MembraneWebRTC.kt
```diff
@@ -35,791 +35,810 @@ import org.webrtc.Logging
...
+    ensureScreencastTrack()
+    localScreencastTrack?.let {
+      removeTrackFromLocalEndpoint(it)
+      localScreencastTrack = null
+    }
+
+    setScreencastTrackState(false)
+    if (calledFromToggleScreencast) {
+      screencastPromise?.resolve(isScreencastOn)
+    }
```
mironiasty, line 671: Why we need this condition? Does resolving promise in disconnect cause some issues?

PR #164, opened 2024-05-06
[Android] Initialize audioSwitchManager using onCreate method.

karkakol:
This PR removes initializeWebRTC method to fix issue with reconnecting to room after leave, it also initalizes audioSwichManager using expo lifecycle methods


mironiasty (approved):
LGTM 🍍

